PR #82, opened 2024-09-14
Reduce periphery image size

neicureuil:
## Periphery Image Size Optimization
The original image was approximately 700 MB, which I believe can be significantly reduced. For comparison, the Portainer agent image is around 200 MB.

To achieve this reduction, I refactored the base image to use Alpine instead of Debian-slim, including only the necessary dependencies.

After that I got an image around 150 MB.

All functionality appears to remain consistent with the previous image.

I have do some tests : 
- Deploy and Remove a stack
- Build an image from a repository
- Prune system images
- Read some containers logs
- Prune volumes

image

### DIFF
--- a/bin/periphery/Dockerfile
+++ b/bin/periphery/Dockerfile
@@ -1,21 +1,24 @@
 # Build Periphery
-FROM rust:1.80.1-bookworm AS builder
+FROM rust:1.80.1-alpine AS builder
 WORKDIR /builder
 COPY . .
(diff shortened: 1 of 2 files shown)


mbecker20 (commented):
Looks great! You are right, no reason for such a large image. The binary itself is just 13 MB.

I'll do some testing as well but I have a feeling this will work fine 😄. And I'll get it merged for upcoming 1.14.3. Thanks for the PR!

PR #89, opened 2024-09-22
fix: check if systemd is available for install

BradLugo:
Adds a check at the beginning of setup-periphery.py to verify if `systemctl` is executable by the current user and if systemd is the init system in use. These changes will inform the user we've decided systemd is unavailable and exits before we attempt to use any systemd functionality, but modify this logic to configure other init systems in the future.

Relates to (but doesn't close) https://github.com/mbecker20/komodo/issues/66.

### DIFF
--- a/scripts/setup-periphery.py
+++ b/scripts/setup-periphery.py
@@ -1,5 +1,6 @@
 import sys
 import os
+import shutil
 import platform
 import json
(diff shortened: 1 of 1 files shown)


mbecker20 (approved):
LGTM. Thanks!

PR #91, opened 2024-09-23
fix: change edit url for docsite

BradLugo:
Changes the default edit url provided by the docusaurus template to the correct komodo url.


mbecker20 (approved):
Thanks :)

Issue #87, opened 2024-09-19
[Feature] Rootless periphery agent


FoxxMD:
When a [`periphery` agent container](https://komo.do/docs/setup/connect-servers#install-the-periphery-agent---container) creates some resources (repos,stacks) in a bind mount these files are currently created as `root`. This makes reading/writing any of these files inconvenient if the user on the host is not root.

The container can be run with a non-root user by specify [`user: "UID:GID"`](https://docs.docker.com/reference/compose-file/services/#user) or equivalent with `docker run` and the agent runs fine. It also creates resources with the correct permissions in this instance. However, the agent **cannot** access `docker.sock` due to it being owned by `root:docker` which is for security, I believe.

Please implement a way to allow control over file/folder permissions for resources touched by the agent, or, implement a way to use the container rootless that can still communicate with docker.

#### Enable SSL/HTTP communication with docker

(My preferred solution)

The crate used for docker communication, [`bollard`](https://github.com/fussybeaver/bollard?tab=readme-ov-file#http), supports SSL/HTTP connections out of the box. The current [`connect_with_local_defaults()`](https://github.com/mbecker20/komodo/blob/main/bin/periphery/src/docker.rs#L35) function could be try-caught (?) with a fallback to `connect_with_http_defaults()` and/or `connect_with_ssl_defaults()` -- both of which use environmental variables to initiate a connection which require no config change for the agent.

This would enable using [docker-socket-proxy](https://github.com/Tecnativa/docker-socket-proxy) for communication which would solve non-root communication with docker.

#### `chown` with UID/GID from ENV

Allow a user to specify `UID` and `GID` in environmental variables and then `chown` files touched by the agent. This would allow the agent to continue to work as `root` while solving file permissions for bind mounts.

mbecker20:
Now supports DOCKER_HOST env variable for use with [docker-socket-proxy](https://github.com/Tecnativa/docker-socket-proxy): https://github.com/mbecker20/komodo/releases/tag/v1.15.10

FoxxMD:
Hi @paulora2405 I wrote a blog post that covers how to do this, some gotchas I discovered, and sample code:

https://blog.foxxmd.dev/posts/migrating-to-komodo/#create-komodo-periphery-agents

willyp713:
[...]

PR #131, opened 2024-10-16
Update docker.rs to allow http docker socket connection

Svagtlys:
Add or_else to allow attempt to connect to docker socket proxy via http if local connection fails, as per #87.

This was tested by adding env var DOCKER_HOST to the compose.env file (picked up by periphery container, which is run as a non-root user without the docker.sock volume defined), creating a new stack in Komodo (using redis image), then deploying. Deployment was successful, however the stack is marked as down in Komodo. I double-checked that the periphery agent is able to view running containers and the logs of those containers by exec-ing into the container and running the commands, so I am unsure where the issue lies.

### DIFF
--- a/bin/periphery/src/docker.rs
+++ b/bin/periphery/src/docker.rs
@@ -31,7 +31,8 @@ impl Default for DockerClient {
   fn default() -> DockerClient {
     DockerClient {
       docker: Docker::connect_with_local_defaults()
-        .expect("failed to connect to docker daemon"),
+        .or_else(|_| Docker::connect_with_http_defaults())
+        .expect("failed to connect to docker daemon with either local or HTTP defaults"),
     }
   }
 }
--- a/bin/periphery/src/docker.rs
+++ b/bin/periphery/src/docker.rs
@@ -30,9 +30,8 @@ pub struct DockerClient {
 impl Default for DockerClient {
   fn default() -> DockerClient {
     DockerClient {
-      docker: Docker::connect_with_local_defaults()
-        .or_else(|_| Docker::connect_with_http_defaults())
-        .expect("failed to connect to docker daemon with either local or HTTP defaults"),
+      docker: Docker::connect_with_defaults()
+        .expect("failed to connect to docker daemon"),
     }
   }
 }


bin/periphery/src/docker.rs
```diff
@@ -31,7 +31,8 @@ impl Default for DockerClient {
   fn default() -> DockerClient {
     DockerClient {
       docker: Docker::connect_with_local_defaults()
```
mbecker20, line 33: What about trying this method?
```rs
Docker::connect_with_defaults()
        .expect("failed to connect to docker daemon"),
```

Svagtlys: I changed that and ran and tested and that works the same as the previous setup. Is it better code practice? I'm not familiar with Rust.

mbecker20: Just a slightly different behavior. 

Doing `connect_with_local_defaults` first may succeed to connect to some local docker socket, even if the user passes `DOCKER_HOST=http://..`, as this method will fallback DOCKER_HOST to `unix:///var/run/docker.sock` if you pass an http address: https://github.com/fussybeaver/bollard/blob/713f07f4fe29bd92a5326379e925504f80fad6d9/src/docker.rs#L906. 

So it would only use `connect_with_http_default` as a fallback if that connection fails. Using `connect_with_defaults` will ensure it will use `DOCKER_HOST` even in the case that a local socket is available, as the user would expect.

Svagtlys: Thanks for the explanation. I couldn't find that function in the bollard docs, but what you're saying does make sense. I'll modify the pr.

mbecker20:
Does the Periphery container show any logs?

mbecker20:
Hey, it does work for me using `Docker::connect_with_defaults()`. I deployed [docker-socket-proxy](https://github.com/Tecnativa/docker-socket-proxy?tab=readme-ov-file) and passed this environment:

```
  POST = 1
  CONTAINERS = 1
  ALLOW_START = 1
  ALLOW_STOP = 1
  ALLOW_RESTARTS = 1
  IMAGES = 1
  INFO = 1
  NETWORKS = 1
  SYSTEM = 1
  VOLUMES = 1
```

I passed env `DOCKER_HOST=tcp://localhost:2375` to Periphery container (ran both the containers with `--network=host`). Everything seems to work. Maybe you can double check your setup?

Svagtlys:
Wow alright, yeah I just had DOCKER_HOST passed as tcp://docker-socket-proxy, adding the port was the trick. Thanks! I really appreciate your feedback on this PR.

PR #134, opened 2024-10-17
feat: Add docs link to topbar

FoxxMD:
Often when I'm in the UI I need to reference the docs but there is no quick way to access them. This PR adds a [Docs](https://komo.do/docs/intro) link in the topbar next to the Version. 

It follows the same media breakpoints as version so it's out of the way on smaller screen sizes.

### DIFF
--- a/frontend/src/components/topbar.tsx
+++ b/frontend/src/components/topbar.tsx
@@ -64,21 +64,34 @@ export const Topbar = () => {
 
         {/* Shortcuts */}
         <div className="flex justify-end items-center gap-2">
-          <MobileDropdown />
-          <OmniSearch setOpen={setOmniOpen} className="lg:hidden" />
-          <Version />
-          <WsStatusIndicator />
-          <KeyboardShortcuts />
-          <TopbarAlerts />
-          <TopbarUpdates />
-          <ThemeToggle />
-          <Logout />
+          <MobileDropdown/>
+          <OmniSearch setOpen={setOmniOpen} className="lg:hidden"/>
+          <div className="flex gap-0">
+            <Docs/>
+            <Version/>
+          </div>
+          <WsStatusIndicator/>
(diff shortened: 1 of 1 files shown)


FoxxMD:
As an aside I think you should also add the docs as the **Website** for the repository

image

My "I need to reference the docs" line of thought is usually:

* Check the UI, can't find a docs link
* Open the repo, **check the sidebar,** no website link there
* Scroll down and scan readme, find "See the docs"

networkydev:
+1, was just about to make an issue for this

frontend/src/components/topbar.tsx
```diff
@@ -41,6 +41,16 @@ import {
 import { Badge } from "@ui/badge";
 import { TopbarAlerts } from "./alert/topbar";
 
+const Docs = () => (<a
```
mbecker20, line 44: Can you move this block of code below the `Topbar` function, above the `Version` function?

frontend/src/components/topbar.tsx
```diff
@@ -66,6 +76,7 @@ export const Topbar = () => {
         <div className="flex justify-end items-center gap-2">
           <MobileDropdown />
           <OmniSearch setOpen={setOmniOpen} className="lg:hidden" />
+          <Docs/>
           <Version />
```
mbecker20, line 80: The styling between Docs and Version has too big a gap, it doesn't look quite right. This is because those Docs and Version Button each have some `padding x`, and the container div has `gap-2`, combined they are too much space.

Wrapping <Docs /> and <Version /> in another div with `gap-0` works:

```
<div className="flex gap-0">
  <Docs />
  <Version />
</div>
```

mbecker20 (changes requested):

mbecker20 (approved):
Looks good thanks

PR #136, opened 2024-10-17
docs: Add development docs

FoxxMD:
This is a WIP for a jumping off point for users trying to develop or contribute to Komodo. The doc I've added is missing a few things and I had a few questions about dependencies and modifying test files:

- [x] `cargo build` with rust stable fails

On my machine (Arch kernel 6.10.9) Trying to run `cargo build` with Rust 1.81 (stable) I get a build error:

```
error: package `aws-sdk-sso v1.45.0` cannot be built because it requires rustc 1.78.0 or newer, while the currently active rustc version is 1.75.0
Either upgrade to rustc 1.78.0 or newer, or use
cargo update aws-sdk-sso@1.45.0 --precise ver
where `ver` is the latest version of `aws-sdk-sso` supporting rustc 1.75.0
```

Are you developing with nightly/unstable Rust? I did not see anything specific in the repo about what Rust version is required.

- [x] UI port for docker testing/development

There is no port exposed in [`test.compose.yaml`](https://github.com/FoxxMD/komodo/blob/0dcffbe9a979055e56c3111dc63734b66f37d278/test.compose.yaml). I imagine this is because it is currently used just to test that the container starts up without error? And exposing the default port might conflict with your existing local (non-docker) instance or something else happening on the komodo build runner.

Easiest change would be to just expose the port but another option would be to use a [service profile](https://docs.docker.com/compose/how-tos/profiles/) and add an additional `core-ui` service with exposed port. This would allow existing behavior to be unchanged but users could run `docker compose -p komodo-dev -f test.compose.yaml --profile ui up -d` to get the service with the port exposed. This would make testing docker builds for UI changes easier without any manual editing of compose files on the users part. Thoughts?

- [x] Order and run requirements for komodo

In the **Local and Docs** section it lists the `run` commands required to get a full Komodo instance up and running. Is the order of these commands correct?

If a user only wants to develop for one part of Komodo (lets say core) can they simply re-build/run one part while leaving everything else up? If so, what are the correct commands/order to rebuild/run for each of these:

* core
* periphery
* core/periphery
* frontend

- [x] vscode `tasks.json`

I realize you are using `runnables-cli` but since the repo is configured for vscode already you would consider also supporting [`tasks.json`](https://code.visualstudio.com/Docs/editor/tasks) to mirror `runfile.toml` so users don't need to install another dependency? If so I can try to get this implemented with the current runfile.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,165 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Run Core",
+            "command": "cargo",
+            "args": [
+                "run",
+                "-p",
+                "komodo_core",
+                "--release"
+            ],
+            "options": {
+                "cwd": "${workspaceFolder}",
+                "env": {
+                    "KOMODO_CONFIG_PATH": "test.core.config.toml"
+                }
+            },
+            "problemMatcher": [
+                "$rustc"
+            ]
+        },
+        {
+            "label": "Build Core",
+            "command": "cargo",
+            "args": [
+                "build",
+                "-p",
+                "komodo_core",
+                "--release"
+            ],
+            "options": {
+                "cwd": "${workspaceFolder}",
+                "env": {
+                    "KOMODO_CONFIG_PATH": "test.core.config.toml"
+                }
+            },
+            "problemMatcher": [
+                "$rustc"
+            ]
+        },
+        {
+            "label": "Run Periphery",
+            "command": "cargo",
+            "args": [
+                "run",
+                "-p",
+                "komodo_periphery",
+                "--release"
+            ],
+            "options": {
+                "cwd": "${workspaceFolder}",
+                "env": {
+                    "KOMODO_CONFIG_PATH": "test.periphery.config.toml"
+                }
+            },
+            "problemMatcher": [
+                "$rustc"
+            ]
+        },
+        {
+            "label": "Build Periphery",
+            "command": "cargo",
+            "args": [
+                "build",
+                "-p",
+                "komodo_periphery",
+                "--release"
+            ],
+            "options": {
+                "cwd": "${workspaceFolder}",
+                "env": {
+                    "KOMODO_CONFIG_PATH": "test.periphery.config.toml"
+                }
+            },
+            "problemMatcher": [
+                "$rustc"
+            ]
+        },
+        {
+            "label": "Run Backend",
+            "dependsOn": [
+                "Run Core",
+                "Run Periphery"
+            ],
+            "problemMatcher": [
+                "$rustc"
+            ]
+        },
+        {
+            "label": "Build TS Client Types",
+            "type": "process",
+            "command": "node",
+            "args": [
+                "./client/core/ts/generate_types.mjs"
+            ],
+            "problemMatcher": []
+        },
+        {
+            "label": "Init TS Client",
+            "type": "shell",
+            "command": "yarn && yarn build && yarn link",
+            "options": {
+                "cwd": "${workspaceFolder}/client/core/ts",
+            },
+            "problemMatcher": []
+        },
+        {
+            "label": "Init Frontend Client",
+            "type": "shell",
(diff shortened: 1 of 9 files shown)


mbecker20:
1. Komodo is developed and released on Rust Stable. It looks like your toolchain is not actually up to date, from the log you posted:
```
error: package `aws-sdk-sso v1.45.0` cannot be built because it requires rustc 1.78.0 or newer, while the currently active rustc version is 1.75.0
Either upgrade to rustc 1.78.0 or newer, or use
```
It looks like your version is `1.75.0`.

2. `test.compose.yaml` does not explicitly core port, I develop using OrbStack which uses the `EXPOSE 9120` in 1bin/core/debian.Dockerfile` and auto DNS + SSL: https://orbstack.dev/. Unfortunately it is macOS only. But this allows me to use the same test file to deploy multiple core without port conflict.

3. You can setup a working current Komodo, then rebuild the parts of the system that has changed and restart it. They don't depend on each other for startup. Just deploy your dev database first (Mongo etc), Core will fail to start without a database.

4. I don't use tasks.json. I am happy to accept a PR for tasks.json, or any other dev tooling that will
    - Help contributors.
    - Doesn't break other tooling.

FoxxMD:
Ok, figured out the rust issue thanks for the rustup hint.

**Exposed Port and Docker**

I've added `expose.compose.yaml` that adds the port mapping to the base `core` service and created another runfile task `test-compose-exposed` that runs this variant so it should break any existing tooling. This has been added to the [Docker](https://github.com/mbecker20/komodo/blob/bce7c80479b3e8e1896111b38ffaeb01dd3b4d1a/docsite/docs/development.md#docker) section of the docs.

**VSCode and Devcontainer**

I've added a [devcontainer](https://containers.dev/) that [extends from the base `test.compose.yaml`](https://github.com/mbecker20/komodo/blob/bce7c80479b3e8e1896111b38ffaeb01dd3b4d1a/.devcontainer/devcontainer.json) to enable "native" development within a container environment that has all the necessary dependencies. The config also runs ferretdb so a user doesn't need to do any setup for that.

A [`tasks.json`](https://github.com/mbecker20/komodo/blob/bce7c80479b3e8e1896111b38ffaeb01dd3b4d1a/.vscode/tasks.json) for [VSCode Tasks](https://code.visualstudio.com/Docs/editor/tasks) implements that same functionality as runfile with additional task dependency/parallel running and tasks for initial setup of frontend (generating types, linking client with yarn, etc...) This is detailed in the [documentation](https://code.visualstudio.com/Docs/editor/tasks) so VSCode users only need to do the following to have a full dev environment with komodo running:

* Open repository in devcontainer
* Run task `Init`
* Run task `Run Komodo`

.vscode/tasks.json
```diff
@@ -0,0 +1,165 @@
...
+            },
+            "problemMatcher": []
+        },
+        {
+            "label": "Init Frontend",
+            "dependsOn": [
+                "Build TS Client Types",
+                "Init TS Client",
+                "Init Frontend Client"
+            ],
```
mbecker20, line 123: I've added one more step here, this is copying the latest built typescript client from `client/core/ts/dist` to `frontend/public/client` folder, so it will be served by the core backend along with the static frontend files.

This is full command I use now to do this (in the `runfile.toml` too)
```
[gen-client]
description = "generates typescript types and build the ts client"
cmd = """
node ./client/core/ts/generate_types.mjs && \
cd ./client/core/ts && yarn build && \
cp -r dist/. ../../../frontend/public/client/."""
```

mbecker20:
This is really awesome @FoxxMD ! Whenever you are ready, you can upgrade this PR from draft and I'll give a final review and merge.

FoxxMD:
Rebased on current `main` HEAD with new tasks for copying the TS client to frontend as well as adding Deno as a dependency for the devcontainer. Tested, Actions as expected AFAIK.

mbecker20 (approved):
LGTM

mbecker20:
The docsite updates are deployed to https://komo.do

PR #282, opened 2025-01-31
feature: use the repo path instead of name in GetLatestCommit

unsync:
I had these annoying warning in my logs : 

```
periphery-1  | 2025-01-31T09:19:01.181038Z  WARN periphery::router: request 9cfd7951-de61-4394-8bda-6c37c3b391a7 | type: GetLatestCommit | error: Repo path is not directory. is it cloned?
periphery-1  | 2025-01-31T09:19:16.026513Z  WARN periphery::router: request 3658378e-5e1a-49de-b7d1-234c78d6dc14 | type: GetLatestCommit | error: Repo path is not directory. is it cloned?
```

I did some digging, forked the repo, added the `repo_path` to the error log and i think i found the issue.

The `GetLatestCommit` command takes the repo name as set in the UI, but since we can override the clone path we can end up is a different path than this : 

```
    let repo_path = periphery_config().repo_dir.join(self.name);
```

So the fix, imho, should be to use the repo `path` instead of the name, since even if we don't override the clone path, there is always a value set : 

```
.request(GetLatestCommit {
        clone_path: repo.config.path.clone(),
      })
```

I'm not sure what is the contributions guidelines in terms of testing, but on my end what i did was to build the docker images, pulled it on my servers and saw i didn't get the warnings anymore.

Feedback welcomed 🙏 

PS: i based this PR on the 1.17 branch, but i have a branch ready with the same change for 1.16

### DIFF
--- a/bin/core/src/monitor/mod.rs
+++ b/bin/core/src/monitor/mod.rs
@@ -263,7 +263,7 @@ pub async fn update_cache_for_server(server: &Server) {
   for repo in repos {
     let (latest_hash, latest_message) = periphery
       .request(GetLatestCommit {
-        name: repo.name.clone(),
+        path: repo.config.path.clone(),
       })
       .await
       .map(|r| (r.hash, r.message))
--- a/bin/periphery/src/api/git.rs
+++ b/bin/periphery/src/api/git.rs
@@ -1,3 +1,4 @@
+use std::path::PathBuf;
 use anyhow::{anyhow, Context};
 use git::GitRes;
 use komodo_client::entities::{update::Log, CloneArgs, LatestCommit};
@@ -16,10 +17,10 @@ impl Resolve<super::Args> for GetLatestCommit {
     self,
     _: &super::Args,
   ) -> serror::Result<LatestCommit> {
-    let repo_path = periphery_config().repo_dir.join(self.name);
+    let repo_path = PathBuf::from(self.path);
     if !repo_path.is_dir() {
       return Err(
-        anyhow!("Repo path is not directory. is it cloned?").into(),
+        anyhow!("Repo path {} is not directory. is it cloned?", repo_path.display().to_string()).into(),
       );
     }
     Ok(git::get_commit_hash_info(&repo_path).await?)
--- a/client/periphery/rs/src/api/git.rs
+++ b/client/periphery/rs/src/api/git.rs
@@ -10,7 +10,7 @@ use serde::{Deserialize, Serialize};
 #[response(LatestCommit)]
 #[error(serror::Error)]
 pub struct GetLatestCommit {
-  pub name: String,
+  pub path: String,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, Resolve)]


unsync:
Since this could introduce a non-backward compatible change for the api and clients, the `path` could be set as optional, and take precedence in `/src/api/git.rs`

unsync:
Here is an attempt to have the path taking precedence over the name : 

https://github.com/mbecker20/komodo/compare/main...unsync:komodo:bug/fixGetLatestCommit?expand=1

mbecker20:
Yeah I think it would make sense to take name and path field like this latest comment, could you make those changes to this PR?

unsync:
@mbecker20 sure, just added a rebased the PR.
I can't see to be able to build the 1.17 branch locally so i can test it.

If i were to fix this, is it safe to run periphery 1.17 with core 1.16, or should i update my whole stack ?

mbecker20:
There shouldn't be any breaking changes to Periphery between 1.16.12 and 1.17.0. But you won't be able to test this change end-to-end without also testing Core 1.17.0 build.

What is the issue you are having building locally?

client/periphery/rs/src/api/git.rs
```diff
@@ -10,7 +10,8 @@ use serde::{Deserialize, Serialize};
 #[response(LatestCommit)]
 #[error(serror::Error)]
 pub struct GetLatestCommit {
-  pub name: String,
+  pub name: Option<String>,
```
mbecker20, line 13: I don't think there is a reason to make `name: Option<String>`. It can be left as `String`, the repo name is still useful on Periphery side in all cases (eg to log the repo name)

unsync:
The build failed because of a cargo error related to this `example` path that wasn't found: 

Screenshot 2025-02-03 at 08 50 15

```
#14 0.297   No such file or directory (os error 2)
#14 ERROR: process "/bin/sh -c cargo build -p komodo_core --release &&   cargo build -p komodo_periphery --release" did not complete successfully: exit code: 101
------
 > [builder 9/9] RUN   cargo build -p komodo_core --release &&   cargo build -p komodo_periphery --release:
0.296 error: failed to load manifest for workspace member `/builder/example/*`
0.296 referenced by workspace at `/builder/Cargo.toml`
0.296 
0.296 Caused by:
0.296   failed to read `/builder/example/*/Cargo.toml`
0.296 
0.296 Caused by:
0.297   No such file or directory (os error 2)
------
```

if i commented it out, the build seem to work.

i'll try it out today with the full 1.17 stack 👍

bin/core/src/monitor/mod.rs
```diff
@@ -263,7 +263,8 @@ pub async fn update_cache_for_server(server: &Server) {
   for repo in repos {
     let (latest_hash, latest_message) = periphery
       .request(GetLatestCommit {
-        name: repo.name.clone(),
+        name: Option::from(repo.name.clone()),
+        path: Option::from(repo.config.path.clone()),
```
mbecker20, line 267: Here instead of `path: Option::from(...)` (which will always be `Some(String)`, should use the `komodo_client::entities::optional_string` helper method, so empty string ("") will become `Option::None` variant.

client/periphery/rs/src/api/git.rs
unsync: @mbecker20 good point, fixed in https://github.com/moghtech/komodo/pull/282/commits/c5ada3fc6fe7756942f0d99f078ab4b8e671e997

bin/core/src/monitor/mod.rs
unsync: @mbecker20 i'm new to rust, so thanks for the tips ^^

Fixed it in https://github.com/moghtech/komodo/pull/282/commits/9518a8623d8b685844e6a0e5210d9ecfa2a9d1b1

bin/periphery/src/api/git.rs
```diff
@@ -16,10 +17,13 @@ impl Resolve<super::Args> for GetLatestCommit {
...
   ) -> serror::Result<LatestCommit> {
-    let repo_path = periphery_config().repo_dir.join(self.name);
+    let repo_path = match self.path {
+      Some(p) => PathBuf::from(p),
+      None => periphery_config().repo_dir.join(self.name)
+    };
     if !repo_path.is_dir() {
       return Err(
-        anyhow!("Repo path is not directory. is it cloned?").into(),
+        anyhow!("Repo path {} is not directory. is it cloned?", repo_path.display().to_string()).into(),
```
mbecker20, line 26: `repo_path.display()` is enough to interpolate, no need for `.to_string()`

mbecker20 (approved):
LGTM after the last nitpick

bin/periphery/src/api/git.rs
unsync: @mbecker20 fixed in https://github.com/moghtech/komodo/pull/282/commits/bbefca8c58946f95c6194d1a8e8dce341cfb3193

unsync:
@mbecker20 i tested it locally with compose : 

Screenshot 2025-02-03 at 09 54 45

```
periphery-1  | 2025-02-03T08:53:15.567604Z  INFO periphery::api::git: GetLatestCommit: /data/periphery/repos/stacks
periphery-1  | 2025-02-03T08:53:15.567630Z  WARN periphery::router: request ae006cd6-d189-44f6-9d75-4402ef068177 | type: GetLatestCommit | error: Repo path /data/periphery/repos/stacks is not directory. is it cloned?
periphery-1  | 2025-02-03T08:53:30.567629Z  INFO periphery::api::git: GetLatestCommit: /data/periphery/repos/stacks
periphery-1  | 2025-02-03T08:53:30.567654Z  WARN periphery::router: request 2ed3eb80-b421-4f16-8880-ea4470816cff | type: GetLatestCommit | error: Repo path /data/periphery/repos/stacks is not directory. is it cloned?
```

I added a log line to see which path was used, and here is before/after removing the custom 
clone path : 

```
### WITH CUSTOM PATH (use the path)
periphery-1  | 2025-02-03T08:55:00.575057Z  INFO periphery::api::git: GetLatestCommit: /data/periphery/repos/stacks

### WITHOUT CUSTOM PATH (use config+name)
periphery-1  | 2025-02-03T08:55:15.572553Z  INFO periphery::api::git: GetLatestCommit: /etc/komodo/repos/test-repo
```

If you have other comments or tests i should run let me know, otherwise i'm passing this PR over to you 🚀

mbecker20:
Nicely done. Thanks for your contribution!

Issue #227, opened 2024-12-17
[Feature] Post Deploy


NeurekaSoftware:
Hey, I see that Komodo has a Pre Deploy option to run shell commands before running docker compose up. I'd like to see a Post Deploy option as well.

My use case is running migrations for GitLab's new [container registry metadata database](https://docs.gitlab.com/17.6/ee/administration/packages/container_registry_metadata_database.html). It [doesn't yet](https://gitlab.com/gitlab-org/omnibus-gitlab/-/issues/8670) run migrations automatically between updates. It seems this would be a great option for Komodo!

sliots:
I also have the same requirement, I want to execute "chmod" command after deployment

mbecker20:
Got it, should be trivial to implement fortunately.

PR #288, opened 2025-02-03
feature: add post_deploy command

unsync:
This PR adds a post_deploy command, similar to the pre_deploy.

### Context
i needed to test a use-case for this, the issue that gave me the idea is here : https://github.com/moghtech/komodo/issues/227

### Questions
ℹ️ what is unclear to me is the expected/wanted behaviour in case of the post_deploy fails (we could say the deployment was ok since compose up ran fine, but we can also fail the whole action)

### Settings
Screenshot 2025-02-03 at 15 41 20

### No failure
Screenshot 2025-02-03 at 15 41 43

### Failure
Screenshot 2025-02-03 at 15 41 35

### DIFF
--- a/bin/periphery/src/compose.rs
+++ b/bin/periphery/src/compose.rs
@@ -388,10 +388,12 @@ pub async fn compose_up(
     log
   };
 
-  // push the compose up command logs to keep the correct order
-  res.logs.push(log.clone());
+  res.deployed = log.success;
 
-  if !stack.config.post_deploy.command.is_empty() {
+  // push the compose up command logs to keep the correct order
+  res.logs.push(log);
+
+  if res.deployed && !stack.config.post_deploy.command.is_empty() {
     let post_deploy_path =
         run_directory.join(&stack.config.post_deploy.path);
     if !stack.config.skip_secret_interp {
@@ -444,8 +446,6 @@ pub async fn compose_up(
       ));
     }
   }
-
-  res.deployed = log.success;
 
   Ok(())
 }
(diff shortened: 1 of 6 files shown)


bin/periphery/src/compose.rs
```diff
@@ -388,8 +388,64 @@ pub async fn compose_up(
     log
   };
 
+  // push the compose up command logs to keep the correct order
+  res.logs.push(log.clone());
```
mbecker20, line 392: Can move `res.deployed = log.success;` before pushing the log, I think it makes sense to keep them together. Then you can remove the `.clone()` on log as well.

mbecker20 (changes requested):

bin/periphery/src/compose.rs
```diff
@@ -388,8 +388,64 @@ pub async fn compose_up(
...
+
+  if !stack.config.post_deploy.command.is_empty() {
```
mbecker20, line 394: Should also check `res.deployed` in this if statement, it shouldn't run the post deploy command if the deploy itself failed.

bin/periphery/src/compose.rs
unsync: Done in https://github.com/moghtech/komodo/pull/288/commits/0b74d80ade0ae98e8f4ead5dfdf615ccda7ff56f

bin/periphery/src/compose.rs
unsync: Good point, also done in https://github.com/moghtech/komodo/pull/288/commits/0b74d80ade0ae98e8f4ead5dfdf615ccda7ff56f

mbecker20 (approved):

mbecker20:
LGTM, thanks for the contribution!

PR #295, opened 2025-02-07
feature: allow docker image text to overflow in table

unsync:
This PR attempt to fix a UI issue on mobile.

### Observations
The image of a service is being truncated and makes it unreadable on mobile : 

#### header view
Screenshot 2025-02-07 at 10 23 18

#### table view
Screenshot 2025-02-07 at 10 23 42

Which is kind of annoying, because you cannot see the actual image name anywhere :) 

### Proposed changes
The proposed changes can fix that, but i'm not sure it is the best approach: 

#### header view
Screenshot 2025-02-07 at 10 24 40

#### table view
Screenshot 2025-02-07 at 10 24 59

ℹ️ It was already breaking the text, but only if your image name had a `-`

### DIFF
--- a/frontend/src/components/util.tsx
+++ b/frontend/src/components/util.tsx
@@ -677,7 +677,7 @@ export const DockerResourceLink = ({
       <Icon server_id={server_id} name={type === "image" ? id : name} />
       <div
         title={name}
-        className="max-w-[200px] lg:max-w-[250px] overflow-hidden overflow-ellipsis"
+        className="max-w-[250px] lg:max-w-[300px] overflow-hidden overflow-ellipsis overflow-breakable"
       >
         {name}
       </div>
--- a/frontend/src/globals.css
+++ b/frontend/src/globals.css
@@ -95,3 +95,7 @@
 .asdfa {
   color: #151b25
 }
+
+.overflow-breakable {
+  overflow-wrap: break-word;
+}


frontend/src/components/util.tsx
```diff
@@ -677,7 +677,7 @@ export const DockerResourceLink = ({
       <Icon server_id={server_id} name={type === "image" ? id : name} />
       <div
         title={name}
-        className="max-w-[200px] lg:max-w-[250px] overflow-hidden overflow-ellipsis"
+        className="max-w-[250px] lg:max-w-[300px] overflow-hidden overflow-ellipsis overflow-breakable"
```
unsync, line 680: the column is declared with 300 size in `frontend/src/components/resources/stack/services.tsx`, maybe it's not useful but we can leverage the extra length.

```
            {
              accessorKey: "container.image",
              size: 300,
              header: ({ column }) => (
                <SortableHeader column={column} title="Image" />
              ),
```

mbecker20:
I'm not sure how I feel about text wrapping like this, as it can be unreadable and looks messy. Currently you can hover over the text to see a popup with the full text contents (builtin browser feature). I think the table UI can be improved in other ways too (resizable columns for example) which can solve the issue better. Thanks as always for the contribution.

unsync:
@mbecker20 while I understand the reasoning, maybe it wasn't clear in the description but there is *already* an overflow, here is a screen of the 1.16 on mobile : 

image

So the PR was actually an attempt to get the same behavior when the image name doesn't contains a dash that allows the line to break.

I wish it could have been fixed without additional css ^^

To me it stays an issue, since it makes it impossible to get some images names on all screens when using a mobile (no hover on mobile):

image

image

image

mbecker20:
Yeah maybe was a bit hasty in closing this. You can use the `break-words` tailwind class instead of adding css: https://tailwindcss.com/docs/word-break. It won't let me reopen the PR so when you have a bit, could you rebase to 1.17.0 and open again?

PR #301, opened 2025-02-09
feature: allow docker image text to overflow in table

unsync:
ℹ️ reopen of https://github.com/moghtech/komodo/pull/295

This PR attempt to fix a UI issue on mobile.

### Observations
The image of a service is being truncated and makes it unreadable on mobile : 

#### header view
Screenshot 2025-02-07 at 10 23 18

#### table view
Screenshot 2025-02-07 at 10 23 42

Which is kind of annoying, because you cannot see the actual image name anywhere :) 

### Proposed changes
The proposed changes can fix that, but i'm not sure it is the best approach: 

#### header view
Screenshot 2025-02-07 at 10 24 40

#### table view
Screenshot 2025-02-07 at 10 24 59

ℹ️ It was already breaking the text, but only if your image name had a `-`

PR #304, opened 2025-02-09
feature: update devcontainer node release

unsync:
I believe the devcontainer node release needs to be aligned with the frontend Dockerfile base image:

```
FROM node:20.12-alpine as builder
```

If not, seems like the devcontainer `Init` task is failing


unsync:
Adding this change to https://github.com/moghtech/komodo/pull/301 as it is frontend related, no need for the extra PR.

PR #312, opened 2025-02-13
feature: improve tables quick actions on mobile

unsync:
Another small but annoying thing on mobile :) 

The quick actions above the table are not responsive, causing the whole page to be "wider" that the screen, not sure how to describe it but the page sort of floats all around:

Screenshot 2025-02-13 at 14 56 15

With this change, on small screens the actions are stacked, so the page is on the correct screen size: 

Screenshot 2025-02-13 at 14 56 02

With wide screens: 

Screenshot 2025-02-13 at 14 56 27

### DIFF
--- a/frontend/src/pages/resources.tsx
+++ b/frontend/src/pages/resources.tsx
@@ -56,7 +56,7 @@ export const Resources = () => {
       actions={<ExportButton targets={targets} />}
     >
       <div className="flex flex-col gap-4">
-        <div className="flex flex-col md:flex-row gap-4 items-center justify-between">
+        <div className="flex flex-wrap gap-4 items-center justify-between">
           <div className="flex gap-4">
             {(is_admin || !disable_non_admin_create) && <Components.New />}
             <Components.GroupActions />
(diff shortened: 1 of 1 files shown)


mbecker20:
Can you rebase these changes on latest 1.17.0 branch? Its showing a lot of changed files

unsync:
@mbecker20 done 🚀

frontend/src/pages/resources.tsx
```diff
@@ -56,7 +56,7 @@ export const Resources = () => {
       actions={<ExportButton targets={targets} />}
     >
       <div className="flex flex-col gap-4">
-        <div className="flex items-center justify-between">
+        <div className="flex flex-col md:flex-row gap4 items-center justify-between">
```
mbecker20, line 59: `gap4` looks like typo, do you mean `gap-4`?

mbecker20: Also, instead of these changes, what if you just add `flex-wrap` class?

unsync: > Also, instead of these changes, what if you just add `flex-wrap` class?

@mbecker20 indeed, that works 👍 

I pushed the changes 

Screenshot 2025-02-14 at 09 54 03

mbecker20 (approved):
LGTM

PR #357, opened 2025-03-11
Fix: amd64 architecture could trigger installation of aarch64 binary

mpkossen:
'arm64' was mistyped as 'amd64', meaning if amd64 was the architecture, it would install the aarch64 binary rather than the x86_64 one. This PR fixes that.


mbecker20 (approved):
Good catch, thanks

PR #358, opened 2025-03-12
Update config.tsx - Adding in better wording for Stack "run_directory" setting.

wlatic:
Changing the word repo to stack and then providing an example of where exactly the run_directory will run on the periphery + stack name will be used.


mbecker20:
Thanks for the PR, this is indeed better. I'm just rebasing and merging to 1.17.0 branch to be included in that release.

PR #386, opened 2025-03-28
Adding a read API endpoint to get container states similar to GetDeploymentsSummary but for all containers that Komodo can see.

alexshore:
I wanted to create a custom homepage widget for Komodo so I could just see what was running/stopped and a total count but an endpoint like `GetDeploymentsSummary` and `GetStacksSummary` didn't exist for all containers. 

At first, I just made a separate python API (https://github.com/alexshore/komodo-containers-api) but thought this could be a good way to up my contributions to open source software so heres my attempt at contributing a `GetDockerContainersSummary` endpoint :)

I'm not too fussed as what counts as what category. For my use-case it's helpful for there really to just be 2 final "states" that the docker containers are categorised into (i.e. just Stopped and Running) but thats a bit more specific to my use-case than to Komodo so right now they're split into Stopped, Running, Unknown and Unhealthy but that can be changed.

I also wasn't sure where in the file to put the new code so I just put it all next to `ListAllDockerContainers` as it performs a similar function.

Hope this is all good and doesn't need too many more changes! Thanks for your help in discord too :)

### DIFF
--- a/bin/core/src/api/read/server.rs
+++ b/bin/core/src/api/read/server.rs
@@ -14,7 +14,9 @@ use komodo_client::{
     ResourceTarget,
     deployment::Deployment,
     docker::{
-      container::{Container, ContainerListItem},
+      container::{
+        Container, ContainerListItem, ContainerStateStatusEnum,
+      },
       image::{Image, ImageHistoryResponseItem},
       network::Network,
       volume::Volume,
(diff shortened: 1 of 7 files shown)


client/core/rs/src/api/read/server.rs
```diff
@@ -294,6 +294,35 @@ pub type ListAllDockerContainersResponse = Vec<ContainerListItem>;
...
+#[response(GetDockerContainersSummaryResponse)]
+#[error(serror::Error)]
+pub struct GetDockerContainersSummary {}
+
+/// Response for [GetDockerContainersSummary]
+#[typeshare]
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+pub struct GetDockerContainersSummaryResponse {
+  /// The total number of Containers
+  pub total: I64,
```
mbecker20, line 313: Use u32 as the number type, in line with the other summary responses eg https://docs.rs/komodo_client/latest/komodo_client/api/read/struct.GetDeploymentsSummaryResponse.html

bin/core/src/api/read/server.rs
```diff
@@ -413,6 +415,45 @@ impl Resolve<ReadArgs> for ListAllDockerContainers {
...
+    self,
+    ReadArgs { user }: &ReadArgs,
+  ) -> serror::Result<GetDockerContainersSummaryResponse> {
+    let servers = resource::list_full_for_user::<Server>(
+      Default::default(),
+      user,
+      &[],
+    )
+    .await
+    .context("failed to get containers from db")?;
```
mbecker20, line 429: `.context("failed to get servers from db")?;`

client/core/rs/src/api/read/server.rs
mbecker20: I actually realized some are using u32, some I64, I think I will change them all to u32

bin/core/src/api/read/server.rs
alexshore: fixed in 9db8273

client/core/rs/src/api/read/server.rs
alexshore: yeah I just used I64 as thats what GetDeploymentsSummary was using

fixed in 73f7e89

alexshore:
for the frontend I don't really have any idea, currently on the dashboard it shows servers, deployments, and stacks, so it doesn't really make much sense to show containers too.

i only really wanted it so an external API could access it, but if we added a server filter param like in `ListDockerContainers` or `ListAllDockerContainers` then you could just have a summary at the top of each server page in the frontend i guess? often you'll have quite a lot of containers running I think so could be good to see statuses directly rather than having to scroll a long list

i know almost nothing about frontend though so I wouldnt even know where to start in adding something like that

mbecker20 (approved):

PR #402, opened 2025-04-09
Improve docs around running periphery in a container

theRAAPster:
Hello, I want to start by thanking you for all the work you have put in to komodo. I just found it a couple weeks ago and have been migrating my systems to it. I have been using systemd to run periphery per the recommendations that you and FoxxMD have made, but I don't think that is practical for the last couple of systems, so I had to dig into running periphery in a container and ran into a couple things that weren't clear to me and required some guess and check, so I'm submitting this PR to hopefully eliminate that uncertainty for people in the future.

Happy to take any feedback and make changes as you see fit.


mbecker20:
Thanks very much for the contribution! Just merging it into 1.17.1 for inclusion in the next release.

PR #404, opened 2025-04-10
Ntfy as Alerter

GFXSpeed:
Adds [ntfy](https://github.com/binwiederhier/ntfy) as an Alerter

grafik

before with custom alerter / after 
grafik


mbecker20:
Looks good! I'm merging into `1.17.1` to be included in this release.

Issue #313, opened 2025-02-13
core & periphery only listen on IPv4, not IPv6


mbunkus:
Both the core & the periphery daemons have their listening address hardcoded at `0.0.0.0`. Only the port number to listen on is configurable. The unfortunate effect is that they only listen on IPv4, not IPv6, no matter whether the host has IPv6 configured or not.

Please add support for listening on IPv6, too, or at least for making the listening address configurable. Note that the equivalent of `0.0.0.0` in IPv4 is `::` in IPv6, which if used together with a port number must be written as `[::]:port` so that the address part `::` can be distinguished from the address/port separator `:`. Please note further that listening on `::` will effectively listen on both IPv6 & IPv4 on some OSes (Linux) while only listening on IPv6 on others (Windows, if I'm not mistaken).

Hardcoded addresses:
  * in core: https://github.com/moghtech/komodo/blob/main/bin/core/src/main.rs#L86
  * in periphery: https://github.com/moghtech/komodo/blob/main/bin/periphery/src/main.rs#L31

Hardcoded IPv6 "listen on every interface, every address" would look like this:

```rs
  let socket_addr =
    SocketAddr::from_str(&format!("[::]:{}", core_config().port))
      .context("failed to parse socket addr")?;
```

Thanks for considering.

vosdev:
Very bummed out to find out Komodo does not listen on IPv6 and can also not be configured to do so. 

Meaning I won't be able to use Komodo as our environment is IPv6 first and only.

It would be a very easy fix, I hope they add this soon.

mbunkus:
Yeah I saw the release notes yesterday. Thank you very much!

PR #405, opened 2025-04-11
Make listener address configurable & Add support for IPv6

vosdev:
This PR makes the listener address configurable for both Core and Periphery through the following env vars:

`CORE_LISTENER_ADDRESS`
`PERIPHERY_LISTENER_ADDRESS`

But the main purpose of this PR is to add support for IPv6.

Before:
```
ss -tulpn | grep 120
tcp   LISTEN 0      1024              0.0.0.0:9120       0.0.0.0:*    users:(("core",pid=4694,fd=13))          
tcp   LISTEN 0      1024              0.0.0.0:8120       0.0.0.0:*    users:(("periphery",pid=4702,fd=10))
```

After:
```
ss -tulpn | grep 120
tcp   LISTEN 0      1024                      *:8120             *:*    users:(("periphery",pid=163028,fd=9))    
tcp   LISTEN 0      1024                      *:9120             *:*    users:(("core",pid=163036,fd=15))
```

Fixes https://github.com/moghtech/komodo/issues/313

### DIFF
--- a/client/core/rs/src/entities/config/core.rs
+++ b/client/core/rs/src/entities/config/core.rs
@@ -44,8 +44,8 @@ pub struct Env {
   pub komodo_host: Option<String>,
   /// Override `port`
   pub komodo_port: Option<u16>,
-  /// Override `listener_address`
-  pub komodo_listener_address: Option<String>,
+  /// Override `bind_ip`
+  pub komodo_bind_ip: Option<String>,
   /// Override `passkey`
   pub komodo_passkey: Option<String>,
   /// Override `passkey` with file
@@ -258,10 +258,10 @@ pub struct CoreConfig {
   #[serde(default = "default_core_port")]
   pub port: u16,
 
-  /// Address the core web server listens on.
+  /// IP address the core server binds to.
   /// Default: [::].
-   #[serde(default = "default_core_listener_address")]
-  pub listener_address: String,
+   #[serde(default = "default_core_bind_ip")]
+  pub bind_ip: String,
 
   /// Sent in auth header with req to periphery.
   /// Should be some secure hash, maybe 20-40 chars.
@@ -528,7 +528,7 @@ fn default_core_port() -> u16 {
   9120
 }
 
-fn default_core_listener_address() -> String {
+fn default_core_bind_ip() -> String {
   "[::]".to_string()
 }
 
@@ -582,7 +582,7 @@ impl CoreConfig {
       title: config.title,
       host: config.host,
       port: config.port,
-      listener_address: config.listener_address,
+      bind_ip: config.bind_ip,
       passkey: empty_or_redacted(&config.passkey),
       first_server: config.first_server,
       frontend_path: config.frontend_path,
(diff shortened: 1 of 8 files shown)


config/core.config.toml
```diff
@@ -33,6 +33,11 @@ host = "https://demo.komo.do"
 ## Default: 9120
 port = 9120
 
+## The address the core system will listen on.
```
mbecker20, line 36:
```
## The address the core server will listen on.
## The default will allow it to accept external IPv4 and IPv6 connections.
```

config/periphery.config.toml
```diff
@@ -18,6 +18,11 @@
 ## Default: 8120
 port = 8120
 
+## The address the server will listen on.
```
mbecker20, line 21:
```
## The address the periphery server will listen on.
## The default will allow it to accept external IPv4 and IPv6 connections.
```

mbecker20 (changes requested):
Thanks for the PR! I've just updated the PR to merge to 1.17.1 branch so I can include it in the next release.

I do have a change to request, I think the name `listener_address` may lead to some confusion about what this does, and may lead to users inputing their servers IP address or something like that. I think my preferred term is `bind_ip`. This more closely matches the mongo configuration for example, they use `bindIp` for this: https://www.mongodb.com/docs/manual/reference/configuration-options/#net-options.

I also left a couple comments clarifying the inline docs.

client/core/rs/src/entities/config/periphery.rs
```diff
@@ -162,6 +164,11 @@ pub struct PeripheryConfig {
   #[serde(default = "default_periphery_port")]
   pub port: u16,
 
+  /// Address periphery listens on.
```
mbecker20, line 167: /// Address periphery server binds to.

client/core/rs/src/entities/config/core.rs
```diff
@@ -256,6 +258,11 @@ pub struct CoreConfig {
   #[serde(default = "default_core_port")]
   pub port: u16,
 
+  /// Address the core web server listens on.
```
mbecker20, line 261: `/// IP address the core server binds to.`

vosdev:
Made the requested changes :)

mbecker20 (approved):
Perfect, thanks!

Fisiu:
@vosdev, why did you make IPv6 as default?
As a result, you may encounter errors:

```
komodo-periphery  | 2025-06-08T14:21:58.222636Z  INFO periphery: Komodo Periphery starting on https://[::]:8120
komodo-periphery  | Error: Address family not supported by protocol (os error 97)
```

```
komodo-core       | 2025-06-08T14:21:55.535186Z  INFO core: Komodo Core starting on http://[::]:9120
komodo-core       | Error: failed to start http server
komodo-core       | 
komodo-core       | Caused by:
komodo-core       |     Address family not supported by protocol (os error 97)
```

With IPv6 as the default, users must adjust their configurations to include the `KOMODO_BIND_IP` and `PERIPHERY_BIND_IP` environment variables. Adding these to the `.env` file is straightforward, but the setup documentation does not cover this.

Issue #414, opened 2025-04-15
Environment variables of stacks and repos flipped


LeadRR:
Since https://github.com/moghtech/komodo/pull/383, the paths of the env vars for the stacks and repos are flipped.

Current state:
PERIPHERY_REPO_DIR=$PERIPHERY_ROOT_DIRECTORY/**stacks**
PERIPHERY_STACK_DIR=$PERIPHERY_ROOT_DIRECTORY/**repos**

Target state:
PERIPHERY_REPO_DIR: ${PERIPHERY_ROOT_DIRECTORY}/**repos**
PERIPHERY_STACK_DIR: ${PERIPHERY_ROOT_DIRECTORY}/**stacks**

https://github.com/moghtech/komodo/blob/main/compose/compose.env

PR #415, opened 2025-04-15
Fix compose.env

LeadRR:
Flip "stacks" and "repos" paths

Fixes https://github.com/moghtech/komodo/issues/414


beckerinj (approved):
Thanks for the catch 🙌

PR #421, opened 2025-04-17
added pushover alerter

alexshore:
added a pushover alerter https://pushover.net/

I am too stupid to know how to add this but it would be nicer if the user didn't have to manually add the URL params themselves and there was instead a UI section that showed if you selected pushover to input the user key, the application token, and possibly any other params (pushover has a lot of other params)


mbecker20 (approved):
LGTM

PR #452, opened 2025-04-27
Fix docker container link

SmoofOperator:


mbecker20:
Thanks for opening, the link you replaced is indeed incorrect. It is supposed to link here though: https://github.com/moghtech/komodo/pkgs/container/komodo-periphery

PR #492, opened 2025-05-07
Removing ServerTemplate in docs

jeroenvds:
Removing ServerTemplate from Resources documentation, as it was removed in Release v1.17.5


mbecker20:
Thanks for the catch.

PR #493, opened 2025-05-07
Ntfy alerter: Add support for email notifications, notification titles and message priorities

FelixBreitweiser:
Originally I only wanted a way to receive email notifications from komodo alerts, and found out that ntfy allows forwarding notifications via email by adding a "X-Email" header to a notification request. So I implemented that, and while I was at it, I also added support for customizing the notification title and message priority:

- in the frontend when setting the alerter type to "ntfy", new configuration fields will pop up
- the default notification title was change from "ntfy alert" to "Komodo Alert", and can be customized if desired
- the main use-case for message priorities is for usage with the ntfy android app, which will notify the user more (or less) urgently (see [the ntfy documentation](https://docs.ntfy.sh/publish/#message-priority))

For the future, maybe it would be nice to have email alert support directly as part of komodo, but that of course entails a bit more work than this:

- email support could maybe be handled with the [lettre](https://crates.io/crates/lettre) crate? 
- smtp server configuration must be handled (as a resource, or just via `core.config.toml`?)
- multiple email recepients could be supported (ntfy only supports one "X-Email" header, so for each email receipient a separate alerter must be configured, which should also be publishing to a different topic. Its all a bit inconvienient)

### DIFF
--- a/bin/core/src/alert/mod.rs
+++ b/bin/core/src/alert/mod.rs
@@ -130,8 +130,8 @@ pub async fn send_alert_to_alerter(
         )
       })
     }
-    AlerterEndpoint::Ntfy(NtfyAlerterEndpoint { url }) => {
-      ntfy::send_alert(url, alert).await.with_context(|| {
+    AlerterEndpoint::Ntfy(NtfyAlerterEndpoint { url, email }) => {
+      ntfy::send_alert(url, email.as_deref(), alert).await.with_context(|| {
         format!(
           "Failed to send alert to ntfy Alerter {}",
           alerter.name
--- a/bin/core/src/alert/ntfy.rs
+++ b/bin/core/src/alert/ntfy.rs
@@ -5,6 +5,7 @@ use super::*;
 #[instrument(level = "debug")]
 pub async fn send_alert(
   url: &str,
+  email: Option<&str>,
   alert: &Alert,
 ) -> anyhow::Result<()> {
   let level = fmt_level(alert.level);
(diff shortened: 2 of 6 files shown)


mbecker20:
Hey, thanks for the PR. I like the addition of email property at the alerter config level, but I'm not sure about the inclusion of `title` and `priority` here as well. Unless I am misunderstanding things, those two come from the alert type being sent, with the title being comparable to eg the Slack title, and the priority coming from the [`SeverityLevel`](https://docs.rs/komodo_client/latest/komodo_client/entities/alert/enum.SeverityLevel.html).

mbecker20:
Also, can you change the target of this PR to the `1.17.6` branch instead of `main`? This way it can smoothly be included in the next release.

FelixBreitweiser:
I've changed the base to `1.17.6`. In regards to the `title` and `priority`: I have purposefully split the PR into two commits, the first of which only adds the email support, so that could be merged on its own. However, the `title` of an Ntfy alert by default is "ntfy alert". Wouldn't something like "Komodo alert" be a bit more descriptive on the receiving end? In any case, getting email support merged on its own would already be quite useful to us.

mbecker20:
I agree that the title and priority should be passed through to each ntfy message, but not that these things should be configured at Alerter level. You could pipe them through from each alert but if you don't want to include that here, you could just remove the second commit and request another review.

FelixBreitweiser:
Okay, I have removed the commit in question. Now only the E-Mail options remain. The `title` still remains "ntfy alert" for all alerts, but that may be a topic for another pull request.

mbecker20 (approved):
LGTM, thanks

shadybraden:
For anyone looking at this later, here is the solution for message priority:
`?p=low`
i.e. tack that on to `endpoint.params.url` like this: `endpoint.params.url = "https://ntfy.domain.com/topic?p=low"`

PR #512, opened 2025-05-13
1.17.6

mbecker20:
# Changelog

🚨 This release moves **official support to FerretDB v2**. Users who deployed v1.17.5 or before using Postgres / Sqlite option are using FerretDB v1 and should eventually migrate using the [FerretDB v2 Update Guide](https://github.com/moghtech/komodo/blob/1.17.6/bin/util/docs/copy-database.md#ferretdb-v2-update-guide). Note that this is not a breaking change to Komodo itself, and users can update to 1.17.6 and continue to use FerretDB v1 if they wish.

🚨 Admins managing user permissions may need to modify the access rules to continue to access features. In particular, **container logs**, **`docker inspect`** on containers, and **terminal access** are now gated behind additional permission (for non admin users)

### Specific Permissions

The main purpose of this release is to refine the access control / permissions system in Komodo. In 1.17.5 and before, access to resources was controlled only via access level (`Read`, `Execute`, `Write`). These levels provide access to the associated `/read`, `/execute`, and `/write` methods on resources, and it worked pretty well to provide RBAC.

Now with more potentially sensitive features, this is not quite enough to provide granular access control. To address this, `specific` permissions have been introduced *in addition* to `Read`, `Execute`, and `Write` levels.

- `Terminal`: User can access the associated resource's terminal.
    - If given on a `Server`, this allows server level terminal access.
    - If given on a `Stack` or `Deployment`, this allows container exec terminal (even without `Terminal` on `Server`)
- `Attach`: User can "attach" *other resources* to the resource.
    - If given on a `Server`, allows users to attach `Stacks` and `Deployments`
    - If given on a `Builder`, allows users to attach `Builds`
- `DockerInspect`:  User can "inspect" docker resources (like containers) on the `Server`
    - Access to this api will expose all container environments on the given server, and can easily lead to secrets being leaked.
- `DockerLogs`: User can retrieve docker / docker compose logs on the associated resource.
    - Valid on `Server`, `Stack`, `Deployment`
    - For admins wanting this permission by default for all users with read permissions, see below on default user groups.
- `ProcessList`: User can retrieve the full running process list on the `Server`

The above `specific` permissions are defined in a list alongside their `level`. This list is open for future expansion / and the associated implementations may be refined in future releases as well.

### Default User Groups

Sometimes you will want to set a "baseline" set of permissions that all users will have on the Komodo instance. Previously this could only be done in very barebones way, by setting `KOMODO_TRANSPARENT_MODE=true` on the Komodo Core container. This would give all users a base level of "Read" on all resources.

In addition to the above permissions features, this release also adds an `everyone` mode to User Groups. If you enable this mode on a User Group, then all users will inherit those permissions as a base.

### Misc.

- **UI**: Fix the inline rename behavior when renaming multiple resources in a row.
- **Alerter**: Ntfy endpoints now support configuring email. Note that you must also make sure SMTP is configured on the Ntfy server. By @FelixBreitweiser in #493

### DIFF
--- a/bin/core/src/api/write/mod.rs
+++ b/bin/core/src/api/write/mod.rs
@@ -69,6 +69,7 @@ pub enum WriteRequest {
   AddUserToUserGroup(AddUserToUserGroup),
   RemoveUserFromUserGroup(RemoveUserFromUserGroup),
   SetUsersInUserGroup(SetUsersInUserGroup),
+  SetEveryoneUserGroup(SetEveryoneUserGroup),
 
   // ==== PERMISSIONS ====
   UpdateUserAdmin(UpdateUserAdmin),
--- a/bin/core/src/api/write/user_group.rs
+++ b/bin/core/src/api/write/user_group.rs
@@ -2,10 +2,7 @@ use std::{collections::HashMap, str::FromStr};
 
 use anyhow::{Context, anyhow};
 use komodo_client::{
-  api::write::{
-    AddUserToUserGroup, CreateUserGroup, DeleteUserGroup,
-    RemoveUserFromUserGroup, RenameUserGroup, SetUsersInUserGroup,
-  },
+  api::write::*,
   entities::{komodo_timestamp, user_group::UserGroup},
 };
 use mungos::{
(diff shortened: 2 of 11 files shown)


fathonyfath:
Hey, I'm really appreciating the ongoing improvements to Komodo! I wanted to flag that requiring a database migration would be a significant breaking change. Users on tagged versions like `1` or `1.17` would likely encounter unexpected issues upon restarting their deployments if a migration becomes necessary.

mbecker20:
@fathonyfath This release marks the move to officially supporting FerretDB v2 (and dropping v1), and includes upgrade guide for this. Users will still be able to upgrade Komodo itself separately and continue using FerretDB v1 for as long as they like, so its not actually breaking.

PR #552, opened 2025-05-26
feat: add musl support for periphery builds

R3D2:
Adds musl-based Docker builds for Komodo Periphery, enabling statically linked binaries that result in ~15MB container images compared to ~100MB+ glibc-based images.

- Uses `rust:1.82-alpine` for modern Cargo.lock compatibility
- Native architecture builds (no cross-compilation complexity)
- Scratch-based runtime for minimal footprint
- Tested builds: AMD64 (16.6MB) and ARM64 (15.7MB)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2544,6 +2544,7 @@ dependencies = [
  "anyhow",
  "async_timing_util",
  "bson",
+ "chrono",
  "clap",
  "derive_builder",
  "derive_default_builder",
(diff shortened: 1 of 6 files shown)


mbecker20:
I appreciate the time you put into this PR. I spent time experimenting with `musl` as well, and found the build to take significantly longer / be less reliable. Additionally, as @FoxxMD said supporting another target will likely result in additional issues related to this, which I am not looking to take on right now.

If you are interested, you may consider creating a community `alpine` distribution for Komodo. Distribution owners are responsible for building and releasing open source software using their own pipelines, and providing distribution-specific support. Users could then point to your images / binaries if they are looking for `musl` support. I am happy to link to it in the docs so long as it fairly stays up to date with official releases and maintains a good relationship with the community.

Issue #567, opened 2025-05-31
Invalid OIDC endpoint causes Rust panic


atakiya:
Unsure if this would be a bug report or feature request or something else entirely..

So if Komodo at startup cannot set up OIDC correctly when enabled per config, a certain code path that calls `unwrap()` causes the entire application to panic.
I do not think Komodo should be designed like this - in my instance I have both GitHub and a self-hosted OIDC provider set up. As long as **at least one** way to authenticate is available, it should simply disable them.
Alternatively, config option to make sure they are all available or vice versa?

Here's the Rust error just for clarity:
```Rust
core-1  |
core-1  | thread 'tokio-runtime-worker' panicked at bin/core/src/auth/oidc/client.rs:54:6:
core-1  | called `Result::unwrap()` on an `Err` value: Failed to initialize OIDC client.
core-1  |
core-1  | Caused by:
core-1  |     0: Failed to get OIDC /.well-known/openid-configuration
core-1  |     1: Server returned invalid response: HTTP status code 400 Bad Request at https://auth.example.com/.well-known/openid-configuration
core-1  | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
core-1  | Error: task 3 panicked with message "called `Result::unwrap()` on an `Err` value: Failed to initialize OIDC client.\n\nCaused by:\n    0: Failed to get OIDC /.well-known/openid-configuration\n    1: Server returned invalid response: HTTP status code 400 Bad Request at https://auth.example.com/.well-known/openid-configuration"
```

For the reason why I encountered this error: I'm moving some subdomains around and currently my OIDC provider only returns 400s. Yes this is on me, but I assumed I could still use Komodo and log in via the GitHub IdP or username+password instead of plucking around in the CLI more.

In the case this is intended design, I'd like to understand why exactly.

mbecker20:
good point, its not necessary to panic in this case. It should just disable the oidc provider as an option and still allow alternatives

Issue #563, opened 2025-05-30
[bug] komodo cannot stop containers because is using wrong container name in command line


Tipiak99:
Hi

Issue with docker stop command from komodo : 

My container name is : `realtime-dev.supabase-realtime`

But Komodo is launching the following command changing the dot with an underscore: 
`docker stop realtime-dev_supabase-realtime`

as a consequence it throw an error : Error response from daemon: No such container: realtime-dev_supabase-realtime

Issue seems to be located only with the stop command, all other commands are using the correct name

thanks!

Tipiak99:
hi,

issue is still there in v18.1

Tipiak99:
@mbecker20

dbswarco:
I'm new to Komodo but just ran across this same issue. In my situation it's case-sensitivity. The container name is something like "Apple", but the command Komodo attempts to run is "docker stop **a**pple && docker container rm **A**pple".

mbecker20:
Hi, it should be fixed in 1.18.2, please confirm and I will close

Tipiak99:
> Hi, it should be fixed in 1.18.2, please confirm and I will close

I confirm that it is now working as expected with the last version.

thanks!!

Issue #557, opened 2025-05-28
bug(oidc): komodo fails to fetch user details from IdP user info endpoint


Crowley723:
Komodo's OIDC authentication implementation only uses claims from the ID token and does not fetch additional user details from the provider's UserInfo endpoint when required claims are missing.

Expected Behavior:
When authenticating via OIDC, users should see their email address and full name in Komodo, regardless of which claims the provider includes in the ID token.

Actual Behavior:
Users authenticated through OIDC providers that don't include email/name/groups claims in the ID token (like Authelia 4.39+) appear with their sub claim (UUID, or other identifier) instead of their email address, and have blank user info fields.

Steps to Reproduce:

    Configure Authelia to act as an OIDC provider.
    Configure Komodo as an OIDC client for Authelia.
    Attempt login with Komodo.
    See UUID instead of email once logged into Komodo.

Additional Information:
I believe this issue is the same one fixed by https://github.com/mealie-recipes/mealie/pull/5228 in Mealie.

Per the [OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims), it is not required for OIDC providers to return anything except for [specific claims](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) in the id_token given to the client after authentication. As such, if the client expects info about the user (such as email, username, etc) they may have to fetch said info from the provider's user info endpoint.

In the recent update for authelia (4.39), we stopped including this information ([to achieve OIDC certification](https://www.authelia.com/blog/we-are-now-openid-certified/)) by default which requires some clients to use the [UserInfo](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) endpoint. We [provided a workaround](https://www.authelia.com/integration/openid-connect/openid-connect-1.0-claims/#restore-functionality-prior-to-claims-parameter) for admins to continue using the old method but this is a band-aid fix.

Possible Solution:
As can be seen in the mealie pr, this should be a fairly straight forward fix, if the returned token doesn't contain the required information about the user, fetch it from the UserInfo endpoint instead.

mbecker20:
Thanks for pointing this out, it should be addressed  in 1.18.2

PR #550, opened 2025-05-26
feat: add maintenance window management to suppress alerts during planned activities

R3D2:
• Implement timezone-aware scheduling with Daily, Weekly, and OneTime maintenance window types
• Create dedicated Maintenance tab in server configuration with DataTable-based management interface

Screenshot 2025-05-26 at 18 49 57
Screenshot 2025-05-26 at 18 50 08
Screenshot 2025-05-26 at 18 50 15

PR #591, opened 2025-06-07
1.18.2

mbecker20:
# Changelog

The two main features of this release are **Linked Repos** and **Alert Maintenance Windows**.

## Repo Linking

For Repo linking, this means you can link Repos that you configure once in Komodo to *multiple resources*. This avoids having to configure your repo provider, name, and account multiple times (you can just select it in the dropdown), and also keeps things in sync if you change the underlying repo provider, name, branch, or account. This is available on all Resources that use git repos (`Stacks`, `Builds`, and `Resource Syncs`).

For Stacks, this means that multiple Stacks attached to the same Repo will **share the same repo clone on individual hosts**. You can migrate your Stacks to used the linked Repo in place, just note that it won't deal with any data mounted inside the clone folders. Those would need to be moved to the new Repo clone path (as specified in the Repo config). Ideally all your data mounts are outside of the repo. Config files which are committed inside the repo are fine to keep mounted in using relative path however.

Note that you can but don't need to attach a Server / Builder to the Repo. If you do configure an attachment on the Repo, the Repo can still be attached to eg Stacks on other Servers. The Repo server / builder attachments aren't considered by Resources depending on the Repo. 

## Alert Maintenance Windows

These allow you to configure time periods to suppress alerts when downtime is expected, such as for system updates. You can schedule either Daily, Weekly, or One Time maintenance windows, and configure them on either individual `Servers`, or on individual `Alerters`.

### Resource

- **Server**: Add docker lists search re #548 
- **Server**: Add Periphery version indicator to Server tables
- **Deployment / All containers**: Support capitals and `.` in the names re #563 
- **Actions**: Add `komodo.execute_container_exec`, `komodo.execute_deployment_exec`, `komodo.execute_stack_exec` methods
    - Similar to `komodo.execute_terminal`, but set up to connect directly to container exec
    - Only requires `disable_container_exec = false` in Periphery config. Direct terminals can be disabled.
- **Stack, Build, Resource Sync**: Can now link **Komodo Repos** (in the sidebar) directly to these resources
    - **Stacks**: Have them share a single repo clone on your hosts.
    - *Don't* need to attach Repo to any Server / Builder, but you can if you would like, to manually pull / reclone.
    - *Note*: Non admin users need `Attach` permission on the Repo to be able to link Resources to it.
- **Server / Alerter**: Add per-server and per-alerter maintenance window configuration to suppress alerts during planned time periods. #550 by @R3D2 and @mbecker20 

### Misc

- Failed OIDC provider initialization just error logs, doesn't crash core. Can still login with other providers. Re. #567 
- OIDC login gets username using the standard `userinfo` endpoint, instead of token claims. Does not affect already created users in Komodo, only new users. Re. #557


undaunt:
#550 is actually by @R3D2 fyi @mbecker20

PR #595, opened 2025-06-10
Add default stack contents configuration option

dylan-hoefsloot:
Adds an option on servers to configure the default content populated in when creating a new stack

### DIFF
--- a/bin/core/src/resource/server.rs
+++ b/bin/core/src/resource/server.rs
@@ -71,6 +71,7 @@ impl super::KomodoResource for Server {
         state: status.map(|s| s.state).unwrap_or_default(),
         region: server.config.region,
         address: server.config.address,
+        default_stack_file_contents: server.config.default_stack_file_contents,
         send_unreachable_alerts: server
           .config
           .send_unreachable_alerts,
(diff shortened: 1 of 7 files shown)


dylan-hoefsloot:
@mbecker20 Perhaps needs some more work, I am not familiar with Rust or Tanstack.

It works for the simple case though.

paulora2405:
This would be really useful, I have a few things that I put in every compose.yaml file, and every new stack I create, I need to copy these common configs from another running stack.

mbecker20:
I appreciate the PR. However, I do have other plans to handle resource defaults across the board using this concept of Template resources. The idea is to mark some particular resources (ie. Stacks) as templates. Then the "New Resource" buttons show any templates you configured in a dropdown, where you can select one as a starting point. This should fill the same use case as this PR does, and be more generic / dynamic (supporting all resource types / configuration fields to default). I don't want to add other default behavior until this can be done.

Until then, users in situation like @paulora2405 can already use the existing Copy resource functionality. So when making new Stack, navigate to the existing Stack with the common configs, and use the Copy button in the upper right. You should never have to manually copy configs over like you say.

PR #618, opened 2025-06-23
[PR] Fix for the spamming /auth/GetUser when unexpected logout occured

huzky-v:
# Description
Environment: 
Komodo core version: 1.18.3
Browser: Chrome / Safari

The case happens when I put the komodo browser idle long enough (e.g. put a sleep on my mac and keep my browser open) , the token timed out. 
I have seen the network and cpu usage goes up. 

After checking the browser console, I found that `/auth/GetUser` are spammed to the Komodo core server.
Here are the reproducible steps:
1. Login Komodo
2.  Deleting `komodo-auth-token` from Local Storage, forcing them to logout, and wait for approx. 10s to redirect back to login page
3. The `/auth/GetUser` requests are getting spammed to the Komodo core server, spiking the CPU and network usage.
spam

And after extensive checking, found that the `qc.invalidateQueries` does not return updated `userData` after the invalidation. (Checked the data shows that the `query.data` still have the user data, which the new response on `/auth/GetUsers` is a HTTP 500 error) 

Later found that the `qc.resetQueries` will stop the spam, but since I do no exp on React, I am not sure if it is best practice or not breaking other imports using `useUser`, breaking existing logics etc.

### DIFF
--- a/frontend/src/lib/hooks.ts
+++ b/frontend/src/lib/hooks.ts
@@ -37,7 +37,7 @@ export const useLoginOptions = () =>
   });
 
 export const useUser = () => {
-  const userInvalidate = useUserInvalidate();
+  const userReset = useUserReset();
   const query = useQuery({
     queryKey: ["GetUser"],
     queryFn: () => komodo_client().auth("GetUser", {}),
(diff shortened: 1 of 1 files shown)


mbecker20 (approved):
I was able to confirm the issue you pointed out re GetUser spam, and confirm that this fixes the issue. Thank you very much!

PR #664, opened 2025-07-14
Implement ability to run actions at startup

bpbradley:
This feature adds a "Run On Startup" feature for Actions. When enabled, the action will be executed shortly after Komodo Core finishes loading.

Now first, I am experienced in C, not Rust, so forgive me if I did anything stupid. I mostly just borrowed from how actions were run in `schedule.rs` for borrowing the syntax / patterns to do what I wanted to do.

Originally, I added the startup actions in `startup.rs` at the end of the `on_startup` function (like this)

```rust
pub async fn on_startup() {
  tokio::join!(
    in_progress_update_cleanup(),
    open_alert_cleanup(),
    ensure_first_server_and_builder(),
    clean_up_server_templates(),
    run_startup_actions()
  );
}
```

And this "worked", but the actions always failed with

```
Stderr
Download http://localhost:9120/client/lib.js
error: Import 'http://localhost:9120/client/lib.js' failed.
    0: error sending request for url (http://localhost:9120/client/lib.js): client error (Connect): tcp connect error: Connection refused (os error 111): Connection refused (os error 111)
    1: client error (Connect)
    2: tcp connect error: Connection refused (os error 111)
    3: Connection refused (os error 111)
    at file:///action-cache/Wy9zMYzrte.ts:1:37
```

My reasoning was that this needed to occur _after_ the server becomes available here

```rust
 if config.ssl_enabled {
    info!("🔒 Core SSL Enabled");
    rustls::crypto::ring::default_provider()
      .install_default()
      .expect("failed to install default rustls CryptoProvider");
    info!("Komodo Core starting on https://{socket_addr}");
    let ssl_config = RustlsConfig::from_pem_file(
      &config.ssl_cert_file,
      &config.ssl_key_file,
    )
    .await
    .context("Invalid ssl cert / key")?;
    axum_server::bind_rustls(socket_addr, ssl_config)
      .serve(app)
      .await
      .context("failed to start https server")
  } else {
    info!("🔓 Core SSL Disabled");
    info!("Komodo Core starting on http://{socket_addr}");
    axum_server::bind(socket_addr)
      .serve(app)
      .await
      .context("failed to start http server")
  }
}
```

However, I don't really know how to get it to run after this in the way I was hoping.

So I sort of hacked it by just spawning the process to run the actions in the background with a delay, before the server starts.

```rust
  tokio::spawn(async {
      // Wait for 5 seconds before running the function to give 
      // server time to startup
      tokio::time::sleep(std::time::Duration::from_secs(5)).await;
      startup::run_startup_actions().await;
  });
```
This now works, and actions execute successfully a few seconds after server starts.

But it definitely feels like I could probably do this in a way that doesn't introduce a possible race condition. That said, I felt 5 seconds was pretty generous.

With that, it does work for me.

image

And the action runs successfully after server startup

image

And actions which do not have this enabled do not run at startup, as expected. So functionally I think it works.

Very welcome to changing how I do this, like I said I am not super familiar with this toolchain.


mbecker20:
Thanks for the PR, could you rebase to 1.18.5 branch and change the PR into that? Then the changes can be merged there and smoothly incorporated into next release.

bpbradley:
Should be all set @mbecker20 I can do the same for my other PR to make it easier for you.

mbecker20:
Hi, hope to get to this shortly, I have been traveling

bpbradley:
> Hi, hope to get to this shortly, I have been traveling

No problem 🙏 take your time

bin/core/src/startup.rs
```diff
@@ -35,6 +39,52 @@ pub async fn on_startup() {
   );
 }
 
+pub async fn run_startup_actions() {
+    let actions = match mungos::find::find_collect::<Action>(
+        &db_client().actions,
+        None,
```
mbecker20, line 45: Can directly query for actions with this config using `doc! { "config.run_at_startup": true }`. Can just make this change myself after merge

mbecker20 (approved):
LGTM, very clean.

Issue #660, opened 2025-07-12
Undefined user agent


cristhianwon:
I have a problem from Komodo with OIDC, I have this IDP protected with a WAF and this block the request from komodo to OIDC because the request does not have a user agent.

A recommended protection from WAF is only accept request with defined user agent.

Now my solution is disable this protection from WAF but I recommend to add the user agent from http client in komodo.

<img width="1576" height="436" alt="Image" src="https://github.com/user-attachments/assets/6a0f1c52-671a-4c9f-8bdc-f97de5a8fea3" />

eleith:
this would also help with general investigation and tuning.

for example, i'm seeing a call to `/application/o/komodo/jwks/` and `/application/o/komodo/.well-known/openid-configuration` every minute!

the current hypothesis is that since the `userAgent` is empty, it's coming from komodo.

it's unclear how to turn this down as 1 minute doesn't correspond to any configurable interval (`MONITORING`, `RESOURCE_POLL` and `JWT_TTL`)

eleith:
i can confirm why and where this happens. 

komodo spawns an oidc check every 60 seconds to keep the jwk refresh here: https://github.com/moghtech/komodo/blob/118ae9b92caf95ae0fdabe9f5984923af4c3bd9d/bin/core/src/auth/oidc/client.rs#L43-L62

this was introduced in https://github.com/moghtech/komodo/pull/248/commits/3f767ed42e328be6887ee0b8b330669d6810dc44

eleith:
setting the user agent might be a small change, according to the [reqwest docs](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.user_agent) the following method:

https://github.com/moghtech/komodo/blob/118ae9b92caf95ae0fdabe9f5984923af4c3bd9d/bin/core/src/auth/oidc/mod.rs#L34-L42

could be modified to call something like

```rust
  static REQWEST: OnceLock<reqwest::Client> = OnceLock::new();
  REQWEST.get_or_init(|| {
    reqwest::Client::builder()
      .redirect(reqwest::redirect::Policy::none())
      .user_agent("Komodo/1.18.4")
      .build()
      .expect("Invalid OIDC reqwest client")
  })
```

cristhianwon:
@eleith maybe the problem is with the reqwest lib where is missing the User-Agent header

[...]

PR #701, opened 2025-08-01
add user agent to oidc requests

eleith:
Komodo is making a request to the oidc discoverable endpoint every 60s.

currently the user agent is empty, making investigations into request utilization difficult. 

this adds a Komodo specific user agent to allow others to easily discover when Komodo is making requests to their other services.

fixes: #660


mbecker20 (approved):
Thanks for the PR

PR #714, opened 2025-08-05
Frontend mobile fixes

dotsam:
A handful of fixes to prevent overflowing the viewport in most places on mobile, and add a little more density in a few places.

I'm unsure what the original intent was of setting up most of the resource Tabs to be grid, but I'm not seeing any regressions when removing it.

### DIFF
--- a/frontend/src/components/layouts.tsx
+++ b/frontend/src/components/layouts.tsx
@@ -185,7 +185,7 @@ export const Section = ({
     {(title || icon || titleRight || titleOther || actions) && (
       <div
         className={cn(
-          "flex flex-wrap gap-2 justify-between",
+          "flex flex-wrap gap-2 justify-between overflow-x-scroll",
           itemsCenterTitleRow ? "items-center" : "items-start"
         )}
       >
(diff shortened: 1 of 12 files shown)


mbecker20:
I actually need to add back the `grid gap-4`, this is important for the spacing. `flex flex-col gap-4` would do similar.

Edit. I don't think its this actually but something seems a bit weird with the border lines
Screenshot 2025-08-07 at 5 12 38 PM 
^ this PR
Screenshot 2025-08-07 at 5 12 56 PM

dotsam:
Hmmm, without the `overflow-x: scroll`, things definitely overflow at mobile size...
Screenshot 2025-08-09 at 12 54 04

And I wasn't seeing the same issue in your screenshot until I checked with Firefox and the macOS "always show scrollbars" setting enabled, so I'll see if I can sort that out, or maybe find a way for these tabs to wrap nicely, but I wasn't having any luck before.

dotsam:
So according to MDN (https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x) `scroll` is supposed to always show scroll bars, but it seems like only Firefox does that. The default for `overflow-x` is actually `visible`, so setting `overflow-x-auto` here seems to actually get the desired behaviour in all browsers.

PR #719, opened 2025-08-07
Add manual network interface configuration for multi-NIC Docker environments

MP-Tool:
# Problem
In Docker setups with multiple network interfaces, users cannot specify which interface should handle internet traffic, leading to unpredictable routing behavior. This causes problems when working with external hosted repositories. While Docker Compose version `2.32+` provides network interface priority `gw_priority` settings (which I recommend using bevor), some edge case scenarios require the ability to manually set the container interface for the default gateway.

# Solution
Added `internet_interface` configuration allows manual specification of which network interface of the container should be used as the default gateway. The feature automatically detects Docker environments and configures the routing table to route internet traffic through the specified interface.

## How it works:

1. Validates the specified interface exists and is UP
2. Discovers the gateway for the interface's network
3. Sets the interface as the default route for internet traffic
4. Provides clear error messages if privileges or configuration are missing

# Usage
```
# config/core.config.toml
internet_interface = "eth1"
```

```
# docker-compose.yaml
services:
  core:
    cap_add: [NET_ADMIN]
    environment:
      KOMODO_INTERNET_INTERFACE: "eth1"
```

# Changes
- New: mod.rs - Network configuration module
- Modified: Core config to support internet_interface field
- Modified: Startup integration for network setup

# Requirements
- `NET_ADMIN` capability for Docker containers
- iproute2 (injected by debian-deps.sh)


bin/core/src/startup.rs
```diff
@@ -27,13 +27,24 @@ use crate::{
...
 pub async fn on_startup() {
+  // Configure manual network interface if specified
+  let config = core_config();
+  if !config.internet_interface.is_empty() {
+    if let Err(e) = network::configure_internet_gateway(
+      config.internet_interface.clone()
+    ).await {
+      tracing::warn!("Failed to configure internet gateway: {e:#?}");
+    }
+  }
```
mbecker20, line 46: You can move `let config = core_config()` and the error handling including log inside fn network::configure_internet_gateway. Then just call `network::configure_internet_gateway().await;` in `on_startup`.

config/core.config.toml
```diff
@@ -89,6 +89,11 @@ repo_directory = "/repo-cache"
 ## Default: /action-cache
 action_directory = "/action-cache"
 
+## Interface to use as default route in multi-NIC environments.
+## Env: KOMODO_INTERNET_INTERFACE
+## Optional, no default.
+# internet_interface = "eth1"
```
mbecker20, line 95:
```suggestion
## Interface to use as default route in multi-NIC environments.
## Env: KOMODO_INTERNET_INTERFACE
## Example: `internet_interface = "eth1"`
## Optional, no default.
internet_interface = ""
```

mbecker20: If you look at the file with linting, it is helpful to have the field not be commented out to help it stand out.

client/core/rs/src/entities/config/core.rs
```diff
@@ -542,6 +549,10 @@ fn default_title() -> String {
   String::from("Komodo")
 }
 
+fn default_internet_interface() -> String {
+  String::new()
+}
```
mbecker20, line 554: Can delete

bin/core/src/network/mod.rs
mbecker20, line 1: If its only one file, no need for folder / mod.rs. Just move this all to `bin/core/src/network.rs`

bin/core/src/network/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+//! It allows Komodo Core to specify which network interface should be used as the default route
+//! for internet traffic, which is particularly useful in complex networking setups with multiple
+//! network interfaces.
+//! 
+//! ## Features
+//! - Automatic container environment detection
+//! - Interface validation (existence and UP state)
+//! - Gateway discovery from routing tables or network configuration
+//! - Safe default route modification with privilege checking
+//! - Comprehensive error handling and logging
```
mbecker20, line 13: Komodo Core uses outbound network for these things
1. Auth providers
2. Git providers
3. Periphery connections

How does setting interface here relate to each of these things? Will it also mean Peripheries are reached via the same interface? I think ideally a user could specify which interface they use for Auth, Git, Periphery individually.

Even if this isn't supported in this PR, I want to make sure this is open for future extension if these features are requested.

mbecker20 (changes requested):
Thanks for the PR, overall it looks good, there's just a few things I have highlighted

bin/core/src/startup.rs
MP-Tool: Absolutely makes sense! 👍

mbecker20 (approved):
LGTM Thanks

PR #732, opened 2025-08-15
Add RunStackService API implementing `docker compose run`

bpbradley:
Adds backend support for `docker compose run` targeting a service defined within a komodo managed stack.

Here is a resource TOML that can be used to create a stack, action, and procedure for testing / validation. Each test is intended to run the same command, which prints `hello, world!` to stdout in a roundabout way (just to demonstrate usage of most of the features -- I also added a random [[TZ]] to demonstrate secrets interpolation was working).

```toml
[[stack]]
name = "helloworld"
tags = ["example"]
[stack.config]
server = "server-E1m5R"
post_deploy.command = """
  # Add multiple commands on new lines. Supports comments.
"""
file_contents = """
services:
  hello_world:
    image: alpine:3.19
    environment:
      - TZ=${TZ}
"""
environment = """
  # VARIABLE = value
TZ=[[TZ]]
"""

##

[[procedure]]
name = "RunStackProcedure"
tags = ["example"]

[[procedure.config.stage]]
name = "Stage 1"
enabled = true
executions = [
  { execution.type = "RunStackService", execution.params.stack = "helloworld", execution.params.service = "hello_world", execution.params.command = ["-lc", "echo \"$ENV1\", \"$ENV2\"!"], execution.params.no_deps = true, execution.params.env.ENV2 = "world", execution.params.env.ENV1 = "hello", execution.params.user = "1000:1000", execution.params.entrypoint = "/bin/sh", enabled = true }
]

##

[[action]]
name = "TestRunStack"
tags = ["example"]
[action.config]
file_contents = """
const update = (await komodo.execute_and_poll("RunStackService", {
  stack: "helloworld",
  service: "hello_world",
  command: [
    "-lc",
    'echo "$ENV1", "$ENV2"!'
  ],
  no_deps: true,
  entrypoint: "/bin/sh",
  env: { 
    ENV1: "hello",
    ENV2: "world"
  },
})) as Types.Update;

const runLog = update.logs.find(log => log.stage === "Compose Run");
if (runLog) {
  if (runLog.success) {
    console.log(runLog.stdout);
  } else {
    console.error(runLog.stderr);
  }
} else {
  console.error("No 'Compose Run' stage found in logs.");
}
"""
```
And km cli implementation

```
❯ km execute run-stack-service --help
Runs a one-time command against a service using `docker compose run`. Response: [Update]

Usage: km execute run-stack-service [OPTIONS] <STACK> <SERVICE> [COMMAND]...

Arguments:
  <STACK>       Id or name
  <SERVICE>     Service to run
  [COMMAND]...  Command and args to pass to the service container

Options:
      --no-tty                   Do not allocate TTY
      --no-deps                  Do not start linked services
      --service-ports            Map service ports to the host
  -e, --env <ENV>                Extra environment variables for the run
      --workdir <WORKDIR>        Working directory inside the container
      --user <USER>              User to run as inside the container
      --entrypoint <ENTRYPOINT>  Override the default entrypoint
      --pull                     Pull the image before running
  -h, --help                     Print help
```
Sample: 

```sh
km execute run-stack-service \
  --no-deps \
  --entrypoint /bin/sh \
  -e 'ENV1=hello&ENV2=world' \
  helloworld hello_world \
  -lc 'echo "$ENV1", "$ENV2"!'
```

### Screenshots

image
image


bpbradley:
After testing, there are some quirks I identified. I have some ideas for them, but figured its a good opportunity to review first and you can decide the correct path forward. 

With the current implementation, the typescript and http API work as intended, and procedure configuration works correctly from what I can tell.

The quirks surround the km cli.

### Multiple env vars in km cli
1I wanted the ability to set multiple extra env vars like so `-e ENV1=hello -e ENV2=world`, but this is unsupported on `HashMap` types, which is detailed here: https://github.com/clap-rs/clap/discussions/4291.
2. The workaround suggested is to use something like Vec<(String, String)> instead with a simple custom parser, but this is not supported by Typeshare and so it's not a good option. There are some other ways to get around this, but its messy (i.e. can have a separate variable that is the tuple, with a directive to skip it for typeshare, then do merging of that variable later). 
3. The alternative I see here is to just do either a String or Vec<String> and handle parsing out the env vars in the actual execution. This makes semantics for the typescript api less clean though (i.e. passing in a Record<string, string> vs just string)

That said, it -does- work with multiple envs as is, which is why I kept it how it is. You just need to do it like `-e 'ENV1=hello&ENV2=world'`

### Argument parser fails in certain unique cases

I might be able to tweak the settings for the command arg, but it seems clap keeps parsing into the command string and so if you start the command with something clap thinks is an option, it will interpret it as so.

For example, with the example I am using, if you use command `-lc echo "$ENV1", "$ENV2"`, it will fail with 

```
`error: unexpected argument '-l' found

  tip: to pass '-l' as a value, use '-- -l'`
  
````

That said, it gives the solution in the tip, i.e. do `-- '-lc echo "$ENV1", "$ENV2"'` instead. I may be able to use `trailing_var_arg(true)` to fix this, since command is the only optional positional argument and is expected at the end of the command.

### Final Notes

1. I did make a bit of an opinionated choice here, as I wasn't sure I wanted to deal with the complications, but I made the api always use the `--rm` flag. Of course, that could easily be an option and even defaulted to true. But I wasn't sure it would be a good idea to have it just spin up containers that ostensibly are intended for one-off commands, and then not clean them up after and just leave them in a stopped state on the host that need to be manually pruned by the user.
2. It may be the case that the user is relying on pre/post deploy steps for their execution, but I did not include that step here as to keep it at least somewhat simple. I thought about including this with another `Option<bool> deploy` which would do the pre/post deploy steps that are typically done with the `DeployStack` command. That said, I think just expecting the user to deploy the stack once first will cover most use cases. I think. I am happy to add that option though if it seems necessary.

File: client/core/rs/src/api/execute/mod.rs [hunk 1]
```diff
@@ -152,6 +152,9 @@ pub enum Execution {
   StopStack(StopStack),
   DestroyStack(DestroyStack),
   BatchDestroyStack(BatchDestroyStack),
+  /// Runs a one-time command against a service using `docker compose run`.
+  #[clap(alias = "run")]
```
mbecker20, line 156: I think this alias is too generic unfortunately.

However for common stuff geared towards stacks, I want to make dedicated `km` command for this

```
#[clap(alias = "compose")]
Stack(stack::Stack)
```

So ultimately this could be more conveniently called at `km compose run ...` in place of the top level Execute alias.

bpbradley: Makes sense, I kind of felt the same. I can just kill the alias for now

mbecker20:
Overall without the 'run' alias, LGTM

client/core/rs/src/api/execute/stack.rs
```diff
@@ -348,6 +350,69 @@ pub struct DestroyStack {
...
+#[empty_traits(KomodoExecuteRequest)]
+#[response(Update)]
+#[error(serror::Error)]
+pub struct RunStackService {
+  /// Id or name
+  pub stack: String,
+  /// Service to run
+  pub service: String,
+  /// Command and args to pass to the service container
+  #[serde(default)]
```
mbecker20, line 374: Don't need serde(default) on Option type

bpbradley: fixed

bpbradley:
I think this handles your requested changes @mbecker20 

One final comment for future in case it comes up --

In retrospect, it may have been better design to use `Vec<String>` for command, rather than just `String`. I originally had it this way, but felt it was messy looking using it that way in typescript API, so I changed it to String.

But I am now realizing that Vec<String> probably allows for closer parity between native `docker compose run` and usage with km cli. Here, you have to encapsulate your entire command in quotes, whereas if we did Vec<String> with `trailing_var_arg = true, num_args = 1..` then you wouldn't need the quotes around the command, as is the case with `docker compose run`

mbecker20:
Its still not merged, do you want to try out using Vec?

bpbradley:
> Its still not merged, do you want to try out using Vec?

Up to you -- I was going to test it out tonight regardless.

mbecker20:
Well to me each have upsides and downsides

- String
    - ✚ Easier UI w/ monaco editors, in line with eg Stack.config.environment. Stores comments.
    - − Multiline string in nested Toml array may cause issues. This isn't an issue for Stack env or Action args, since they are top level fields on config.
    - − Worse CLI experience
    
- Vec<String>
    - − Slightly harder UI. Needs to strip comments.
    - ✚ Toml better
    - ✚ CLI better
    
I think Vec<String> has the best tradeoffs if you are ok with switching

bpbradley:
> Well to me each have upsides and downsides
> 
> - String
>     - ✚ Easier UI w/ monaco editors, in line with eg Stack.config.environment. Stores comments.
>     - − Multiline string in nested Toml array may cause issues. This isn't an issue for Stack env or Action args, since they are top level fields on config.
>     - − Worse CLI experience
>     
> - Vec<String>
>     - − Slightly harder UI. Needs to strip comments.
>     - ✚ Toml better
>     - ✚ CLI better
>     
> I think Vec<String> has the best tradeoffs if you are ok with switching

Sounds good. The only thing I'm not sure on is how I should handle the command field in the procedure config, I could have the user arbitrarily add more text boxes etc, but there isn't really room for that. I could have another Monaco editor maybe? But then how should I parse it into the Vec? Splitting on space is sufficient in general but a quote encapsulated string counts as one arg I believe from compose run perspective

mbecker20:
Why does command field need changing?

mbecker20:
I think 1.19.0 is ready, I'm going to go ahead with this release and we can rebase this onto the next release branch

bpbradley:
> Why does command field need changing?

It might not. I'm just not immediately sure how the text from the command field is going to be mapped to the underlying Vec<String> by default. I need to just play around with it in front of me.

bpbradley:
Actually maybe I'm overthinking it because I have control over how it is joined back together in the final execution. Yeah I think it will be a trivial fix.

mbecker20:
Ok, I'll give you an hour before I merge 1.19.0 😈

bpbradley:
> Ok, I'll give you an hour before I merge 1.19.0 😈

Haha id definitely beat that but I got a baby to put to bed 🤣 hence the scrambled brain right now.

Don't wait on me, do your thing! I'll probably get started in an hour or so.

mbecker20:
That's what I thought, then you said trivial fix. I'll go ahead with the release then and set up the next release branch

mbecker20:
Since I squashed 1.19.0 on merge to main, you will have to rebase before i can change the target branch, could use cherry pick

bpbradley:
> Since I squashed 1.19.0 on merge to main, you will have to rebase before i can change the target branch, could use cherry pick

Sounds good, I just finished up the implementation and it seems to work now with this as desired

```sh
km execute run-stack-service \
  --no-deps \
  --entrypoint /bin/sh \
  -e 'ENV1=hello&ENV2=world' \
  helloworld hello_world \
  -lc 'echo "$ENV1", "$ENV2"!'
```

which is exactly how docker compose run works. And you are able to set the command in config in the same text box with `-lc 'echo "$ENV1", "$ENV2"!'`, so also parity with how docker compose run works. Only change is in typescript API, now expects an array.

Ill figure out the rebase to 1.19.1 in a bit, just gonna test some more.

bpbradley:
@mbecker20 should be rebased onto 1.19.1 now, with the target branch changed as well. I updated the PR body to reflect the recent changes as well for clarity / future onlookers.

Now that we have time, anything else needed on this? Any docs updates needed? I assume most of that is handled automatically.

bpbradley:
I actually have some further improvements. I got into some of the weeds I was thinking about earlier with properly parsing / structuring the command in the various methods between Action, Procedure, and CLI. Now I think all of them have parity with how native docker behaves.

The remaining annoyance I had was that I needed to do something like this in Actions

```
  command: [
    "-lc",
    "'echo \"$ENV1\", \"$ENV2\"!'"
  ],
```

To manually handle the correct escaping, whereas in procedure config and km cli, it worked as expected with how docker run works.

But in docker compose, you would be able to do `command:  ["-lc",  'echo "$$ENV1", "$$ENV2"!']` , and so I was hoping I could get that to work without also disrupting procedure/cli which were already working.

Basically I just needed to properly handle shell parsing. I had to add `shell-escape = "0.1.5"` crate for backend and `"shell-quote": "1.8.1"` for frontend for dealing with procedure config.

But now the other features remain working in parity with `docker compose run` and I can also do the much cleaner

```ts
const update = (await komodo.execute_and_poll("RunStackService", {
  stack: "helloworld",
  service: "hello_world",
  command: [
    "-lc",
    'echo "$ENV1", "$ENV2"!'
  ],
  no_deps: true,
  entrypoint: "/bin/sh",
  env: { 
    ENV1: "hello",
    ENV2: "world"
  },
})) as Types.Update;
```
in Action, which has parity with how commands work in docker in the exec form.

If you dont want the added complexity though, no problem. I can revert this commit.

mbecker20:
> But in docker compose, you would be able to do command:  ["-lc",  'echo "$$ENV1", "$$ENV2"!'] , and so I was hoping I could get that to work without also disrupting procedure/cli which were already working.
> Basically I just needed to properly handle shell parsing. I had to add shell-escape = "0.1.5" crate for backend and "shell-quote": "1.8.1" for frontend for dealing with procedure config.

This is very interesting, I think smarter handling of command like this is also necessary in the `exec_terminal`, `exec_container_terminal` and the other like methods. This PR looks good to me, and I appreciate the care you are taking on this. Just need to rebase again on 1.19.1, I had to push some hot fixes to main and already rebased 1.19.1 on this, and let me know when you are ready for it to be merged.

bpbradley:
> > But in docker compose, you would be able to do command:  ["-lc",  'echo "$$ENV1", "$$ENV2"!'] , and so I was hoping I could get that to work without also disrupting procedure/cli which were already working.
> > Basically I just needed to properly handle shell parsing. I had to add shell-escape = "0.1.5" crate for backend and "shell-quote": "1.8.1" for frontend for dealing with procedure config.
> 
> This is very interesting, I think smarter handling of command like this is also necessary in the `exec_terminal`, `exec_container_terminal` and the other like methods. This PR looks good to me, and I appreciate the care you are taking on this. Just need to rebase again on 1.19.1, I had to push some hot fixes to main and already rebased 1.19.1 on this, and let me know when you are ready for it to be merged.

Sounds good! I'm actually traveling for work at the moment but can probably find time to rebase again tomorrow, if not Thursday for sure.

bpbradley:
@mbecker20 I believe we are all set now, just rebased.

PR #766, opened 2025-08-25
Add option to make run command detachable

bpbradley:
I missed this use case in original testing, but it may occasionally be necessary to run the command in a detached state, in the event of a long running command or one where periphery may be disconnected during run.

The only downside of running detached, which is why I didn't originally include it, is that you don't have access to the logs. For this reason, it defaults to disabled (previous behavior). And if you need the logs, you can get them and follow them in the same way by the process ID returned by the run command, which is printed to stdout.

So for example in an Action,

```ts
const update = (await komodo.execute_and_poll("RunStackService", {
....
})) as Types.Update;

const pid = update.logs.find(log => log.stage === "Compose Run").stdout;
```

And then you would be able to create a terminal on that server, and execute in terminal something like `docker logs -f {pid}` to stream the logs.

### DIFF
--- a/bin/core/src/api/execute/stack.rs
+++ b/bin/core/src/api/execute/stack.rs
@@ -1033,6 +1033,7 @@ impl Resolve<ExecuteArgs> for RunStackService {
         command: self.command,
         no_tty: self.no_tty,
         no_deps: self.no_deps,
+        detach: self.detach,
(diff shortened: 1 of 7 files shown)


mbecker20 (approved):
Clean

PR #796, opened 2025-09-01
Add server name filter to Containers tab

### LINKED ISSUE #687: Filtering containers of a server in the Containers tab
In the Containers tab, it would be nice to be able to filter containers by server name, in addition to their name.

Thank you for the awesome work!

antoniosarro:
Resolves #687

## Description

Added the ability to filter containers by server name in addition to existing container name search.

## Changes

- Enhanced existing search input with debounce function to reduce unnecessary computations
- Added multi-select component for filtering containers by server name
- Integrated skeleton loading component for better loading states
- Added reset button that appears when at least one server is selected

## Screenshots
image
image
image
image

### DIFF
--- a/frontend/src/ui/multi-select.tsx
+++ b/frontend/src/ui/multi-select.tsx
@@ -1,0 +1,161 @@
+import * as React from "react";
+import { Check, ChevronsUpDown, X } from "lucide-react";
+import { cn } from "@/lib/utils";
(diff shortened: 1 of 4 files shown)


frontend/src/ui/multi-select.tsx
mbecker20, line 1: This is a helpful component, I think I have some things doing this manually right now but worse

PR #805, opened 2025-09-04
Improve KomodoClient experience for URL and Error handling

FoxxMD:
This PR improves the developer experience for using the  `komodo_client` npm package:

* Compensate for URLs with trailing slashes when joining api path
* Implement a `KomodoApiError` class extending `Error`, used for api response rejection instead of plain object
  * Uses Error Cause for non-komodo errors
  * Error message is extracted from Serror, if present
  * Appends trace, if present, to stack string
  * Exposes response and SError as class fields

Many third party logging and network-wrapping libraries are used to handling rejected/thrown `Error` instances, so using a real `Error` means developers and end users get better readability and visibility into errors for free.

### DIFF
--- a/client/core/ts/src/lib.ts
+++ b/client/core/ts/src/lib.ts
@@ -50,6 +50,25 @@ export type ClientState = {
   secret: string | undefined;
 };
 
+/**
+ * Joins all interim parts with one separator character
+ * 
+ * @see https://stackoverflow.com/a/55142565/1469797
+ */
+const pathJoin = (parts: string[], sep?: string) => {
+    const separator = sep || '/';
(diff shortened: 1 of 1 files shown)


FoxxMD:
The only thing I haven't accounted for here is if Komodo _itself_ (through Action) expects the rejected object to be structured the old way. If this is the case I can add fields to the class to make its structure mirror the old way so Komodo can handle it without changing any other code.

mbecker20:
In `bin/core/src/api/execute/action.rs`, line 309 the Action does handle the catch case in hardcoded way. I think this change may affect this part and must be tested, could you include some test case / results?

Issue #771, opened 2025-08-26
CreateStack POST disables the create button in the "New Stack" modal but then never un-disables it.


Kickbut101:
Komodo Agent/Periphery - 1.19.1
Server - 1.19.1
Firefox - 142.0

I'm sure I'm using the wrong verbage, sorry. 

I noticed that if you go to create a new stack in the UI and the stack happens to have the same name as another stack the "Create" button gets the 'disabled=""' attribute added to it in the HTML. 

The problem is that the attribute doesn't get removed nor the modal refreshed when the user goes to type in a new name. You are forced to refresh the entire page before that button becomes accessible again.

Screenshots:
### Before:
<img width="872" height="653" alt="Image" src="https://github.com/user-attachments/assets/517805cd-ab6e-4a1b-a7dc-ef8d9c948abd" />

### After:
<img width="971" height="667" alt="Image" src="https://github.com/user-attachments/assets/b5d1071e-3c9c-4560-8437-43bd3e76ffd3" />

### New Modal opened:
<img width="871" height="580" alt="Image" src="https://github.com/user-attachments/assets/a117fd28-1170-489e-9c35-2802ecc0f72c" />

antoniosarro:
If this issue hasn't been addressed by anyone, I can check it myself

MP-Tool:
Issue can be closed.

PR #819, opened 2025-09-06
Fix Submit Dialog Button Behavior with 500 Errors on Duplicate Names

MP-Tool:
## Overview
Resolves issue #771 by implementing proper HTTP status codes and improving error handling for resource creation/copying operations.

## Problem
Previously, duplicate name errors returned HTTP 500 (Internal Server Error) instead of appropriate status codes, causing:
- Submit dialogs to close immediately on validation errors
- Poor user experience requiring data re-entry
- Inconsistent error handling across the application
- Console errors instead of appropriate warnings for expected user errors

### Backend Changes
- **Centralized error handling**: Added `handle_resource_creation_error()` function in `bin/core/src/api/write/mod.rs`
- **Proper HTTP status codes**:
  - `409 Conflict` for duplicate names and busy resources
  - `400 Bad Request` for validation errors (empty names, invalid ObjectIds, permission errors)
  - `500 Internal Server Error` only for actual system errors
- **Consistent implementation**: Applied to all resource types (Stack, Server, Action, Repo, Build, Alerter, ResourceSync)
- **Enhanced error messages**: "Resource with name '{}' already exists" for better clarity

### Frontend Changes
- **Status-code-based error detection**: Replaced message-based error parsing with HTTP status code classification
- **Improved dialog behavior**:
  - Create/Copy dialogs remain open on validation errors (409/400)
  - Dialogs close only on successful operations or system errors (500+)
- **Enhanced logging strategy**:
  - Validation errors logged as warnings (expected user behavior)
  - System errors logged as errors (unexpected failures)
- **User-friendly error messages**: Clear, actionable feedback for validation errors

## Impact
- **Better UX**: Users no longer need to re-enter data when fixing validation errors
- **Proper HTTP semantics**: Status codes now correctly reflect the type of error
- **Improved debugging**: Clear distinction between validation issues and system failures
- **Consistent error handling**: Unified approach across all resource creation/copying operations

### DIFF
--- a/bin/core/src/api/write/mod.rs
+++ b/bin/core/src/api/write/mod.rs
@@ -2,7 +2,7 @@ use std::time::Instant;
 
 use anyhow::Context;
 use axum::{
-  Extension, Router, extract::Path, middleware, routing::post,
+  Extension, Router, extract::Path, middleware, routing::post, http::StatusCode,
 };
 use derive_variants::{EnumVariants, ExtractVariant};
 use komodo_client::{api::write::*, entities::user::User};
@@ -10,7 +10,7 @@ use resolver_api::Resolve;
 use response::Response;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
-use serror::Json;
+use serror::{Json, AddStatusCodeError};
 use typeshare::typeshare;
 use uuid::Uuid;
 
@@ -39,6 +39,26 @@ mod variable;
 
 pub struct WriteArgs {
   pub user: User,
+}
+
+/// Helper function to handle resource creation errors with proper HTTP status codes
+pub fn handle_resource_creation_error(e: anyhow::Error) -> serror::Error {
+  let error_msg = e.to_string();
+  
+  // Check if this is a validation error and set appropriate status code
+  if error_msg.contains("Resource with name") && error_msg.contains("already exists") {
+    e.status_code(StatusCode::CONFLICT)
+  } else if error_msg.contains("Must provide unique name for resource") {
+    e.status_code(StatusCode::CONFLICT)
+  } else if error_msg.contains("busy") {
+    e.status_code(StatusCode::CONFLICT)
+  } else if error_msg.contains("Valid ObjectIds cannot be used as names") 
+         || error_msg.contains("Must provide non-empty name") 
+         || error_msg.contains("User does not have permissions") {
+    e.status_code(StatusCode::BAD_REQUEST)
+  } else {
+    e.into()
+  }
 }
 
 #[typeshare]
--- a/bin/core/src/api/write/mod.rs
+++ b/bin/core/src/api/write/mod.rs
@@ -2,7 +2,7 @@ use std::time::Instant;
 
 use anyhow::Context;
 use axum::{
-  Extension, Router, extract::Path, middleware, routing::post, http::StatusCode,
+  Extension, Router, extract::Path, middleware, routing::post,
 };
 use derive_variants::{EnumVariants, ExtractVariant};
 use komodo_client::{api::write::*, entities::user::User};
@@ -10,7 +10,7 @@ use resolver_api::Resolve;
 use response::Response;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
-use serror::{Json, AddStatusCodeError};
+use serror::Json;
 use typeshare::typeshare;
 use uuid::Uuid;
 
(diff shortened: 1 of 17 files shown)


bin/core/src/api/write/mod.rs
```diff
@@ -41,6 +41,26 @@ pub struct WriteArgs {
...
+  } else if error_msg.contains("busy") {
+    e.status_code(StatusCode::CONFLICT)
+  } else if error_msg.contains("Valid ObjectIds cannot be used as names") 
+         || error_msg.contains("Must provide non-empty name") 
+         || error_msg.contains("User does not have permissions") {
+    e.status_code(StatusCode::BAD_REQUEST)
+  } else {
+    e.into()
+  }
+}
```
mbecker20, line 62: I think this is the right idea, but not the best approach. This error handling should be done inline inside `resource::create` to avoid matching on error strings.

Can make `resource::create(...) -> serror::Result<...>` directly to accomplish this.  You can `.status_code(...)` directly on most Result, similar to anyhow's `.context()`.

Note that you may have to fix other usage of `resource::create` due to change in error type, but you can easily convert the serror::Error to anyhow::Error using `.map_err(|e| e.error)`.

frontend/src/components/resources/common.tsx
```diff
@@ -439,12 +439,26 @@ export const CopyResource = ({
...
+      },
+      onError: (error: any) => {
+        // Keep dialog open for validation errors (409 Conflict, 400 Bad Request)
+        // System errors (500, etc.) will close the dialog
+        if (!error.isValidationError) {
+          setOpen(false);
+        }
+        // The standard error handling (toast, logging) is already handled by the useWrite hook
+      },
+    });
```
mbecker20, line 460: Not sure the point of this refactor? Just move this stuff inline above.

frontend/src/components/resources/common.tsx
```diff
@@ -525,11 +538,23 @@ export const NewResource = ({
...
+      return;
+    }
+    try {
+      const result = templateId
+        ? await copy({ name, id: templateId })
+        : await create({ name, config });
+      const id = result._id?.$oid;
+      if (id) {
+        nav(`/${usableResourcePath(type)}/${id}`);
+      }
```
mbecker20, line 552: Why try catch this?

frontend/src/components/resources/common.tsx
```diff
@@ -547,7 +572,10 @@ export const NewResource = ({
           onKeyDown={(e) => {
             if (!name) return;
             if (e.key === "Enter") {
-              onConfirm();
+              onConfirm().catch(() => {
+                // Error is already handled by the useWrite hook and onConfirm
+                // Just prevent uncaught promise rejection
+              });
```
mbecker20, line 578: This seems redundant? Why throw error above just to catch it here?

frontend/src/lib/hooks.ts
```diff
@@ -243,19 +267,41 @@ export const useWrite = <
...
-      if (detail) {
-        msg_log += detail + " | ";
+      
+      // Use appropriate messages for different error types
+      let toastTitle: string;
+      let toastDescription: string;
+      
+      if (e.isValidationError) {
+        toastTitle = "Invalid Input";
+        toastDescription = msg || "Please check your data and try again.";
```
mbecker20, line 292: msg won't ever be falsy

frontend/src/lib/hooks.ts
```diff
@@ -243,19 +267,41 @@ export const useWrite = <
...
+      } else {
+        let msg_log = msg ? msg[0].toUpperCase() + msg.slice(1) + " | " : "";
+        if (detail) {
+          msg_log += detail + " | ";
+        }
+        toastTitle = `Write request ${type} failed`;
+        toastDescription = `${msg_log}See console for details`;
```
mbecker20, line 299: I don't think these changes are necessary, it should be kept close to the api error handling and better error messages should be handled by good top level  `.context()` message on the backend.

frontend/src/components/layouts.tsx
```diff
@@ -254,8 +254,16 @@ export const NewLayout = ({
...
+              } catch (error: any) {
+                // Check if this is a validation error based on status code
+                const status = error?.status || error?.response?.status;
+                const isValidationError = status === 409 || status === 400;
+                
+                // Keep dialog open for validation errors, close for system errors
+                if (!isValidationError) {
+                  set(false);
+                }
+                // Note: Error handling (toast, logging) is handled by useWrite hook
```
mbecker20, line 266: Not sure this change is necessary

mbecker20: Already in the catch case, set(false) is not called

frontend/src/components/resources/common.tsx
mbecker20, line 578: This change not needed if you don't throw error above

Issue #829, opened 2025-09-09
Login Form Doesn't Submit on Enter Key Press


MP-Tool:
Just noticed a small UX hiccup on the login page. When users type in their username and password, hitting the Enter key doesn't submit the form – they have to click the login button manually. This feels a bit clunky and could frustrate folks in a hurry. Let's add a quick event listener to handle the Enter key for a smoother experience. Thanks! 🚀

jackra1n:
will fix that

PR #830, opened 2025-09-09
Allow signing in by pressing enter

jackra1n:
Solves #829 
Signup still works because it still has the `onClick=...`

✅ tested login
✅ tested signup new user
✅ tested login by pressing enter
✅ tested login by pressing enter with wrong login credentials

### DIFF
--- a/frontend/src/pages/login.tsx
+++ b/frontend/src/pages/login.tsx
@@ -112,6 +112,11 @@ export default function Login() {
     login(creds);
   };
   
+  const handleSubmit = (e: any) => {
+    e.preventDefault();
+    handleLogin();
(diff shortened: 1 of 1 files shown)


mbecker20 (approved):

frontend/src/pages/login.tsx
```diff
@@ -185,6 +190,7 @@ export default function Login() {
           {options?.local && (
             <form
               ref={formRef}
+              onSubmit={handleSubmit}
```
mbecker20, line 193: I would have defined handleSubmit function here anonymously instead of named const above, but this difference is pedantic.

PR #933, opened 2025-10-22
added file path when getting compose logs

### LINKED ISSUE #932: Stack on second machine running podman can't be managed in the UI
So I just installed periphery on a bazzite mini-pc and when i try to deploy a stack i get this error

`
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
>>>> Executing external compose provider "/usr/bin/podman-compose". Please see podman-compose(1) for how to disable this message. <<<<

CRITICAL:podman_compose:no compose.yaml, docker-compose.yml or container-compose.yml file found, pass files with -f
Error: executing /usr/bin/podman-compose -p ollama logs --tail 100: exit status 255

`

this issue appears to be that podman-compose requires the compose file to run

as this correctly runs
`
bash-5.2# /usr/bin/podman-compose -f /etc/komodo/stacks/ollama/compose.yaml -p ollama logs --tail 100
5137d2561464 time=2025-10-22T17:48:00.206Z level=INFO source=routes.go:1511 msg="server config"
`

my suggestion would be to add a reference to the stack compose file to the podman-compose commands

nathn123:
attempt to fix https://github.com/moghtech/komodo/issues/932

### DIFF
--- a/bin/core/src/api/read/stack.rs
+++ b/bin/core/src/api/read/stack.rs
@@ -95,6 +95,7 @@ impl Resolve<ReadArgs> for GetStackLog {
         services,
         tail,
         timestamps,
+        compose_files: stack.compose_file_paths().to_vec(),
       })
       .await
       .context("Failed to get stack log from periphery")?;
@@ -130,6 +131,7 @@ impl Resolve<ReadArgs> for SearchStackLog {
         combinator,
         invert,
         timestamps,
+        compose_files: stack.compose_file_paths().to_vec(),
       })
       .await
       .context("Failed to search stack log from periphery")?;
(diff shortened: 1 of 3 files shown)


mbecker20:
Currently these changes would be breaking, it must work with more than one / custom compose file path

mbecker20:
I just don't like that this switches out working behavior for potentially breaking behavior for many users using docker.

How about a new periphery config flag `podman_mode = true` (`PERIPHERY_PODMAN_MODE`) which enables using compose file paths for logs, instead of `-p project-name`? Then podman hosts could be directed to enabled this, while docker hosts would have no change to the working behavior.

PR #955, opened 2025-11-03
Add track option for Additional Env Files

ChanningHe:
## Problem

When using Additional Env Files with Stack resources, Komodo always attempts to read and validate these files. This causes issues when files don't exist in the repository until deployment time (e.g., files decrypted by SOPS to `/dev/shm/` during Pre Deploy).

The UI would throw errors trying to read non-existent files, and potentially interfere with git operations.

## Solution

Added an optional `track` field to `AdditionalEnvFile` configuration:
- When `track: true` (default): Komodo manages the file
- When `track: false`: File is only passed to docker compose via `--env-file`, no UI management

## Notes

I'm not very familiar with Rust, and this codebase has significant complexity. If there are any issues, potential breakages, or better approaches to solve this problem, please let me know - feedback and suggestions are very welcome!

## Testing

- ✅ Backward compatibility: Old string format configs load correctly
- ✅ New format: Track option works as expected in UI
- ✅ Untracked files: Not included in `all_file_paths()` or `all_file_dependencies()`
- ✅ Docker compose: All files (tracked + untracked) passed via `--env-file`


mbecker20:
Good idea, is it possible for you to rebase the changes onto 2.0.0 branch?

client/core/rs/src/entities/stack.rs
```diff
@@ -804,6 +808,97 @@ pub enum StackFileRequires {
...
+impl AdditionalEnvFile {
+  /// Get the file path
+  pub fn path(&self) -> &str {
+    &self.path
+  }
+
+  /// Check if this file should be tracked by Komodo
+  pub fn tracked(&self) -> bool {
+    self.track
+  }
```
mbecker20: This method isn't needed

ChanningHe:
Rebased onto 2.0.0 (base branch updated).

Issue #956, opened 2025-11-03
[Bug] Failed to redeploy stack using webhook


vuongtlt13:
### Description
When sending a GitLab webhook to redeploy a stack, the webhook handler rejects the request because the branch in the webhook payload (master) does not match the branch stored for the stack in the DB (main). The stack is linked to a repository whose branch is actually master, but the stack record shows main, causing the deploy to fail.

### Steps to reproduce

- Create a stack (id: 8347293847293847f) linked to a repository whose branch is **master**.
- Trigger the GitLab webhook for the stack: POST /listener/gitlab/stack/8347293847293847f/deploy (webhook branch/ref = master)

#### Observed behavior

- Webhook handling fails with this log/warning: 2025-11-03T06:02:15.938055Z WARN StackWebhook{id="8347293847293847f"}: core::listener::router: Failed at running webhook for stack 8347293847293847f | request branch does not match expected

- DB: stack 8347293847293847f has branch = main.

- The linked repo expects branch = master.

### Expected behavior

- The webhook should trigger a redeploy when the webhook branch/ref matches the repository branch.
- If the stack record is out-of-sync with the repository (for example due to a branch rename from master ⇄ main), Komodo should:
- Resolve and accept the repository’s current default branch when processing a webhook (and optionally update the stack record), or
- Return a clear error that includes both expected and received branch names and guidance to fix it, or
- Allow branch-aliasing/mappings (e.g., treat master and main as equivalents if configured).

### Root cause:
- Stack branch do not change when updated **linked_repo**

### Solution:
- Update logic updating stack with linked_repo and change git info (git_provider, branch, git_https)
- PR: https://github.com/moghtech/komodo/pull/957

#### Local test:
- Branch change from **main** to **master**  when updating linked_repo
- Before 
<img width="609" height="324" alt="Image" src="https://github.com/user-attachments/assets/fbf1e93b-8832-482e-9f58-de2b0122127f" />

- After

<img width="630" height="570" alt="Image" src="https://github.com/user-attachments/assets/8565595b-4e5b-481a-88b4-06874e11868a" />

SergeyFilippov:
I have exactly the same issue... Webhooks are rejected, if the branch is not "main" (as it being reset in TOML on using Repo instead of direct git connection). 
So, to make web hooks work (and GitOps in general), one can't use Repos with non "main" branch. This is sad 😟.

Delota:
I have the same with Github webhooks and using the build step.
[...]

PR #957, opened 2025-11-03
feat(core/stack): sync linked repo git settings into Stack config

vuongtlt13:
Issue: https://github.com/moghtech/komodo/issues/956

- When setting linked_repo in UpdateStack validation, also sync:
  - git_provider
  - branch
  - git_https

Ensures Stack webhook branch checks and git operations align with the linked Repo configuration.

### DIFF
--- a/bin/core/src/resource/stack.rs
+++ b/bin/core/src/resource/stack.rs
@@ -403,6 +403,9 @@ async fn validate_config(
     .context("Cannot attach Repo to this Stack")?;
     // in case it comes in as name
     config.linked_repo = Some(repo.id);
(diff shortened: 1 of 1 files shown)


mbecker20:
This method will lead to issue because it has doubled "source of truth" for this configuration -- It would be stored in both repo config and stack config, and update to repo config must also update stack config... would be messy.

I've already implemented fix for Stack webhook when using linked repo, it should instead be done in the webhook handler,  in `2.0.0` branch: https://github.com/moghtech/komodo/blob/559ce103da46bb32349231a2a8e6e75ec6737925/bin/core/src/listener/resources.rs#L344

PR #966, opened 2025-11-07
Upgrade Docker base images from Debian Bullseye to Bookworm

MP-Tool:
# Security: Upgrade Docker base images from Debian Bullseye to Bookworm

## Problem
I stumbled upon critical CVEs in the Debian Bullseye base images currently used across all Dockerfiles.

## Solution
Updated all Docker base images from Debian Bullseye to Bookworm (Debian 12):
- `rust:1.89.0-bullseye` → `rust:1.89.0-bookworm`
- `debian:bullseye-slim` → `debian:bookworm-slim`  
- `lukemathwalker/cargo-chef:latest-rust-1.89.0-bullseye` → `latest-rust-1.89.0-bookworm`

## Changes
- **10 Dockerfiles updated** (Periphery, Core, CLI, Binaries)
- **Rust version unchanged** at 1.89.0 to avoid dependency issues
- Only OS layer updated for security patches

## Testing
✅ Successfully built and tested `komodo-periphery:1.19.5` & `komodo-core:1.19.5` with Bookworm
✅ All dependencies compatible with Debian 12
✅ No breaking changes in functionality

## Note
This PR addresses security vulnerabilities. Feel free to use it if it fits your release cycle - no pressure!

---

**Risk:** Minimal - only OS security updates, no Rust version changes

### DIFF
--- a/.devcontainer/dev.compose.yaml
+++ b/.devcontainer/dev.compose.yaml
@@ -1,6 +1,6 @@
 services:
   dev:
-    image: mcr.microsoft.com/devcontainers/rust:1-1-bullseye
+    image: mcr.microsoft.com/devcontainers/rust:1-1-bookworm
     volumes:
       # Mount the root folder that contains .git
       - ../:/workspace:cached
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,7 +3,7 @@
 {
 	"name": "Komodo",
 	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-	//"image": "mcr.microsoft.com/devcontainers/rust:1-1-bullseye",
+	//"image": "mcr.microsoft.com/devcontainers/rust:1-1-bookworm",
 	"dockerComposeFile": ["dev.compose.yaml"],
 	"workspaceFolder": "/workspace",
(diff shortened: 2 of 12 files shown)


Lenart12:
If we are updating might as well update it to trixie instead of bookworm.

MP-Tool:
> If we are updating might as well update it to trixie instead of bookworm.

I chosed Bookworm because Trixie was released only a few months ago. Bookworm currently has the more reliable LTS releases. In addition, Bookworm has support until 30 June 2028.

mbecker20:
In 2.0.0 branch and upcoming v2 release, the distributed images are now trixie based.
The binaries are built in bookworm, building in older environment links to older libc increasing compatibility. For example if it was built in trixie the binary wouldn't run on bookworm or any linux distro with slightly older libc.

lordpayder:
I’m currently planning to take a closer look at Komodo. One of the first things that stood out to me negatively was that the Docker image is more than six months old. It is of course good to see that work is underway on a new major release, but that should not come at the expense of maintaining the current release.

Trivy reports almost 300 security issues in the current image ghcr.io/moghtech/komodo-core:latest, which is still referenced in the documentation, including 5 critical and 27 high-severity findings. This issue could likely be resolved quite easily by simply updating the base image.

PR #889, opened 2025-09-28
2.0.0

mbecker20:
- 🚨 Both **Core** and **Periphery** must be upgraded to v2 for the connection to work.
- 🚨 Action users: **`komodo.execute_terminal` / `komodo.execute_container_exec`** changes -- see **Terminals** section below.

# Changelog

I'm very excited to bring you Komodo v2. The major improvements are:
- **New UI** with higher contrast and better UI primitives
- Add **`Docker Swarm`** integration
- Enable "reversing" the Periphery connection.
- Easy upgrade to public key authentication.
- Implement Server onboarding flow. 
- Improve the Terminal feature set.
- Add passkey and totp 2fa support
- Single user can link multiple primary logins.
- Add full OpenAPI documentation

## **Periphery → Core Connection**  🦎

One of the major limitations in Komodo v1 was that Core always had to establish the connection to Periphery. In many setups, this was not the optimal connection strategy, and would often require the usage of a VPN to establish the connection securely.

Now with v2, **Periphery can be configured to establish an outbound connection to Core**, and communication with Core will be multiplexed over this connection. #281

**Existing Core → Periphery connections also continue working, and no configuration changes are required to update from Komodo v1.**

Ideally Komodo Core should be served over HTTPS with certificates recognized by Periphery, however it also works for Periphery to connect to Core over `ws://...`, as well as insecure TLS (untrusted certificates) if you set `PERIPHERY_CORE_TLS_INSECURE_SKIP_VERIFY=true`.

**It is usually both easier and more secure to use this Periphery → Core connection method**, and there isn't much other difference or downside. The default compose configuration has changed to this method, and I recommend most users migrate their setups when convenient.

### periphery.config.toml
```toml
## The address of Komodo Core. Must be reachable from this host.
## If provided, Periphery will operate in outbound mode.
## Use `ws://12.34.56.78:9120` if you don't have TLS enabled Core.
## Env: PERIPHERY_CORE_ADDRESS
## Default: None
core_address = "example.komodo.com"

## The Server this Periphery agent should connect as.
## If matching Server doesn't exist yet, see "Server Onboarding".
## Using `$(hostname)` will use the output of shell `hostname` command.
## Env: PERIPHERY_CONNECT_AS
## Default: None
connect_as = "$(hostname)"
```

If using a reverse proxy with Komodo Core, you can limit the IPs which can connect to the Periphery endpoint. For example with Caddy:

```caddyfile
(reject-ips) {
[...]

PR #1033, opened 2025-12-12
Standardize naming convention for "passkey." Some variables were plural, some were singular which can lead to confusion.

rrik:


mbecker20:
The reason is that core can only every take a **single passkey**, while periphery can take **multiple passkeys**.
Upcoming v2 also deprecates passkeys for public / private keys, and is configured automatically using onboarding keys / confirming public keys in UI. #889

PR #1040, opened 2025-12-14
Add git-lfs to core and periphery

### LINKED ISSUE #996: Git LFS isn't installed in periphery
I recently setup a build with one of my repos, which utilizes [GIT-LFS](https://git-lfs.com/) to handle large files. The build kept failing until I docker exec'ed into the container and ran the following commands:

```sh
apt update
apt install git-lfs
git lfs install
```

I think adding this block `bin/periphery/debian-deps.sh` should resolve the issue, but I'm not familiar with the codebase.

For now I am using a fairly hacky workaround in the docker-compose config to get things to work:
```yaml
periphery:
  image: ghcr.io/moghtech/komodo-periphery
  # ...
  command: bash -c 'apt update && apt install git-lfs && git lfs install && periphery'
  # ...
```

I really appreciate all the great work you guys are putting into this project, thank you!

NiXTheDev:
> Support large(r) files by adding git-lfs to both core and periphery. Core probably doesn't need this, but just in case since it also installs git

Closes #996


lillithrosepup:
Is there any update on this? It has been 1 major and 3 minors, I see zero reason for this simple PR to be ignored for this long.

NiXTheDev:
> Is there any update on this? It has been 1 major and 3 minors, I see zero reason for this simple PR to be ignored for this long.

I'd want to know too, perhaps we should ask @mbecker20

PR #1053, opened 2025-12-20
feat: Add SOPS and Age support to Periphery

Nikhil-Gohil:
Adding sops and age binaries to the runtime dependencies. This allows users to decrypt secrets (e.g., .env files) directly within the container for GitOps workflows, removing the need for pre-decryption scripts on the host. Will make it much easier to use sops + age based secret handling in komodo. 

Right now, we have to build docker images for core/periphery (tedious) or bind mount these binaries to the docker containers (could break). 

I tested this workflow by creating a komodo periphery image and running a stack:
Repo with periphery + sops/age is here: [Komodo Periphery Sops](https://github.com/Nikhil-Gohil/komodo-periphery-sops)

Can use this as a pre deploy for stacks in komodo assuming secrets are stored as .env.enc in git:
`sops --decrypt --input-type yaml --output-type dotenv .env.enc > .env`

One issue I can think about is this creates a slight difference between komodo periphery as a docker container and the systemd agent since it will be expected that if installing as a systemd agent, sops and age needs to be manually installed.

### DIFF
--- a/bin/core/debian-deps.sh
+++ b/bin/core/debian-deps.sh
@@ -3,7 +3,13 @@
 ## Core deps installer
 
 apt-get update
-apt-get install -y git curl ca-certificates iproute2
+apt-get install -y git curl ca-certificates iproute2 age
+
+# Sops only available as a binary
+SOPS_VERSION="v3.11.0"
+ARCH=$(dpkg --print-architecture)
+curl -L "https://github.com/getsops/sops/releases/download/${SOPS_VERSION}/sops-${SOPS_VERSION}.linux.${ARCH}" -o /usr/local/bin/sops
+chmod +x /usr/local/bin/sops
 
 rm -rf /var/lib/apt/lists/*
 
--- a/bin/core/debian-deps.sh
+++ b/bin/core/debian-deps.sh
@@ -6,7 +6,7 @@ apt-get update
 apt-get install -y git curl ca-certificates iproute2 age
 
 # Sops only available as a binary
-SOPS_VERSION="v3.11.0"
+SOPS_VERSION="v3.12.2"
 ARCH=$(dpkg --print-architecture)
 curl -L "https://github.com/getsops/sops/releases/download/${SOPS_VERSION}/sops-${SOPS_VERSION}.linux.${ARCH}" -o /usr/local/bin/sops
 chmod +x /usr/local/bin/sops
--- a/bin/core/debian-deps.sh
+++ b/bin/core/debian-deps.sh
@@ -3,13 +3,7 @@
 ## Core deps installer
 
 apt-get update
-apt-get install -y git curl ca-certificates iproute2 age
-
-# Sops only available as a binary
-SOPS_VERSION="v3.12.2"
-ARCH=$(dpkg --print-architecture)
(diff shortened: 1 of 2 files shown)


NiXTheDev:
> One issue I can think about is this creates a slight difference between komodo periphery as a docker container and the systemd agent since it will be expected that if installing as a systemd agent, sops and age needs to be manually installed.

That can be handled in the periphery install script, that also needs a disclaimer, and different distros support because apt, dnf, pacman and others I very likely missed have different ways to install packages
```bash
apt/dnf install -y <package> #Oh lol, same syntax
pacman -Sy <package>
```

Or alternatively it can still have the pre-decryption script as a fallback if sops and age aren't installed, and then be slowly phased out(?), emitting a warning to the log about the requirement at first

Bhuemann:
I built docker images to include the SOPS + Age binaries, and found that pre-deployment hooks run after a check is performed ensuring all files exist. This means if you use alternative names for .env a dummy file must exist to pass the check before the SOPS command can run

Nikhil-Gohil:
Realized that sops and age are only needed on the Periphery agents. Updated the PR to remove them from the Core and bumped the Periphery SOPS version to latest.

flowcool:
+1 on native SOPS support in Periphery. Sharing our custom-image approach as another data point, complementary to @Nikhil-Gohil's fork.

**Dockerfile** (both parents pinned by digest so builds cannot drift silently):

```dockerfile
FROM ghcr.io/getsops/sops:v3.13.3@sha256:857f5a151ac0b2bfc55c1e4e5581d66fb8e268e4d106b38e74191f3bac9d58ea AS sops

FROM ghcr.io/moghtech/komodo-periphery:2.2.0@sha256:7fb1a4807d125ce036a17d37c940b4001402afcaf342a2c720c98d096b1b54da

COPY --from=sops /usr/local/bin/sops /usr/local/bin/sops

# Fail the build rather than ship an image whose sops is broken or missing.
RUN sops --version --disable-version-check
```

A tiny GHA reads the upstream version from the Dockerfile and rebuilds on every Periphery bump.

**Notable design choices**:

- **No `age` binary**. `sops` implements age natively as a library and reads the private key via `SOPS_AGE_KEY_FILE`. The `age` CLI is only needed to *generate* keys, which happens off-host. One less binary in the image, one less supply-chain item. Verified: decryption works with `age` absent.
- **Age key mounted read-only** from the host into the Periphery container — it never enters Komodo Core and is never a Komodo variable. Keeps the trust boundary clean.
- **Decryption via `compose_cmd_wrapper`**: `sops exec-env <path> '[[COMPOSE_COMMAND]]'`. Secrets are decrypted only at deploy time, into env vars scoped to the compose process — never materialized as a `.env` file on disk. This also sidesteps the pre-deploy file-check ordering issue tracked in #1234.

Running this on 4 hosts (~30 stacks, ~4 months) with zero regressions. Happy to answer if any of this is useful for shaping the native design.

Pirulax:
Hello,
I'm a new-comer to Komodo, and this would be a feature I'd use.
I'm keeping my `stack.toml` inside my repos that I deploy, I want to keep as much as possible in there.
Obviously putting the env without encryption into a repo isn't a good idea, so this would come handy, instead of having to manually keep sops updated.

PR #1060, opened 2025-12-22
Add the server to Komodo via the Core API / UI.

sky150:
I saw that there was only documentation on how to install the Periphery agent on the server. So I added instructions for adding a server to Komodo via UI and API.


mbecker20:
v2 and onboarding keys no longer require manual server creation #889

PR #1083, opened 2026-01-02
feat: allow selecting docker registry accounts from all registries

### LINKED ISSUE #1058: [Bug] Unable to use provider other than docker.io in deployment
### Description
Custom Docker registries (e.g., ghcr.io) are missing from the Deployment Image account selection, even though they are correctly configured.
But it's visible in the Stack Image Registry settings.

### Steps to Reproduce

Add a non-docker.io registry (e.g., domain="ghcr.io") in the configuration.

Try to set the Image > Account for a deployment.

### Expected Behavior
The configured ghcr.io account should show up in the account list.

### Actual Behavior
The list is empty and no accounts can be selected.

Image

Image

carl-ac:
- Modified AccountSelector to show accounts from all docker registries (not filtered by single provider)
- Added domain labels to docker accounts (e.g., 'username (ghcr.io)')
- Maintains backward compatibility with git provider behavior
- Enables users to select accounts from ghcr.io, docker.io, etc. in Deployment config

Resolves #1058

### DIFF
--- a/frontend/src/components/config/util.tsx
+++ b/frontend/src/components/config/util.tsx
@@ -837,26 +837,47 @@ export const AccountSelector = ({
       : ["ListDockerRegistryAccounts", "ListDockerRegistriesFromConfig"];
   const config_params =
     type === "None" ? {} : { target: id ? { type, id } : undefined };
-  const db_accounts = useRead(db_request, {}).data?.filter(
-    (account) => account.domain === provider
-  );
-  const config_providers = useRead(config_request, config_params).data?.filter(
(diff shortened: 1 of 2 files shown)


mbecker20:
Currently, the account selector will provide the available accounts on relevant provider after the user enters the image (for example `gchr.io/...`). Usually user enters the image first, then selects account, which works to choose intended account. Why do you need to choose account before entering image?

carl-ac:
I think this workflow is fine with me.
Thanks.

PR #1122, opened 2026-01-26
Fix: E0716 temporary value dropped while borrowed

alien43:
Updated the AlertData::Custom formatting logic to resolve a compiler error introduced in newer Rust versions (1.92+). The previous use of format_args! within an if/else block created temporary values that were dropped before they could be used by the outer format! macro.

### DIFF
--- a/bin/core/src/alert/discord.rs
+++ b/bin/core/src/alert/discord.rs
@@ -230,14 +230,13 @@ pub async fn send_alert(
       )
     }
     AlertData::Custom { message, details } => {
-      format!(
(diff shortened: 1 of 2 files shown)


alien43:
I am sorry, I have no idea as I have never written Rust in my life. I just
generated this fix with Gemini in order to get my server running and
decided to upstream it.

azban:
Gotcha. I deleted that comment because I found out that pinning the version sets a minimum, rather than allowing to pin to a specific older version.

mbecker20:
this issue is fixed in v2

PR #1124, opened 2026-01-28
feat: add compose_cmd_wrapper_include for selective command wrapping

ChanningHe:
## Summary
Previously, compose_cmd_wrapper only wrapped 'docker compose up'. This caused
issues when using sops or similar tools that need to wrap other commands like
'config' - the placeholder in additional_env_files would fail before the
wrapper was applied.

- Add `compose_cmd_wrapper_include` configuration to control which docker compose subcommands are wrapped
- Support wrapping
  - **Compose**: `config`, `build`, `pull`, `up`, `run`
  - **Stack**: `config`, `deploy`
- Fix wrapper placeholder text not displaying in MonacoEditor.

## Config
``` Toml
[[stack]]
name = "test"
# ...
additional_env_files = [{ path = "{}", track = false }]
compose_cmd_wrapper = "sops exec-file --no-fifo encrypted.env '[[COMPOSE_COMMAND]]'"
compose_cmd_wrapper_include = ["config", "pull", "up"]
```
## Frontend Screenshots
wp

wp-stack


ChanningHe:
@mbecker20 
Hi, I’m wondering if this PR could be reviewed and considered for merging into the `2.0.0`.
Although the change looks like an extension (adding more wrapped subcommands),
in practice it behaves more like a fix for cases where compose commands were executed before the wrapper was applied.

In my setup, even when using  
`sops exec-env encrypted.env '[[COMPOSE_COMMAND]]'`  
instead of the `sops exec-file --no-fifo` example mentioned in the PR, the issue still exists.

When `env_files` are not involved, inserting variables work fine as long as they are only used under `environment`.
However, once variables are referenced in other sections such as `ports`, deploy fails.This happens because those fields need variables to be resolved during the pull / parse stage.Since the pull stage was not previously wrapped by `compose_cmd_wrapper`, the variables were not available at that point.

### Example
Compose File:
``` compose
services:
  aria2:
    container_name: aria2
    ports:
      - ${ARIA2_LISTEN_PORT}:${ARIA2_LISTEN_PORT}
      - ${ARIA2_LISTEN_PORT}:${ARIA2_LISTEN_PORT}/udp
      - 16800:16800
    environment:
      - RPC_PORT=${ARIA2_RPC_PORT}
      - LISTEN_PORT=${ARIA2_LISTEN_PORT}
      - RPC_SECRET=${ARIA2_RPC_SECRET}
    volumes:
      - $DOCKER_DATA/aria2:/config
      - $PATH_MEDIA/downloads:/downloads
    image: p3terx/aria2-pro:latest
```
Error: `invalid proto`
Snipaste_2026-02-02_12-45-36

ChanningHe:
Rebased onto the latest `2.0.0` branch.

mbecker20:
@ChanningHe this is a nice feature, thanks for keeping it up to date. Its just quite a bit to review and test so appreciate the patience

Issue #808, opened 2025-09-05
Feature Request: Ability to run multiline commands in pre- and post deploy


marrobHD:
Hi,

currently only single line commands work in pre- and post deploy commands.

A fix would be to save & execute Pre Deploy and Post Deploy as scripts. This would fix the limitation.

Example:

```
echo "hello world" 

cat > hi <<- EOM
    hello
    world
EOM
```

This will fail with `sh: 1: Syntax error: "&&" unexpected`

PR #1217, opened 2026-02-24
let user specify if they want command split into multiple lines

ericls:
It's a bit confusing when I write a command and it doesn't work. 
Turns out the spliting behavior is not doing what I want it to do.
So I want a way to turn it off.
I think it make sense to provide user the ability to turn it off.

Screenshot 2026-02-24 at 09 23 10
Screenshot 2026-02-24 at 09 23 14

Relates to but not entirely the same ask from #808

### DIFF
--- a/bin/core/src/alert/discord.rs
+++ b/bin/core/src/alert/discord.rs
@@ -230,14 +230,12 @@ pub async fn send_alert(
       )
     }
     AlertData::Custom { message, details } => {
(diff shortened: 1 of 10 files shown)


ericls:
Should I port this to 2.0.0 instead?

ericls:
Replaced by https://github.com/moghtech/komodo/pull/1265

PR #1238, opened 2026-03-10
fix dev container issues on 2.0.0

sloria:
fixes this error when installing frontend deps in the dev container:

```
error @vitejs/plugin-react@5.1.4: The engine "node" is incompatible with this module. Expected version "^20.19.0 || >=22.12.0". Got "20.12.2"    
```

also addresses an OOM when running `yarn build` via the `Run Komodo` task:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory       
```

### DIFF
--- a/.devcontainer/dev.compose.yaml
+++ b/.devcontainer/dev.compose.yaml
@@ -12,6 +12,7 @@ services:
     ports:
       - "9121:9121"
     environment:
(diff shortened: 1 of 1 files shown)


sloria:
~~actually, this is only needed on the 2.0.0 branch. will rebase~~

edit: done

PR #1240, opened 2026-03-11
update local dev setup instructions

sloria:
updates the "Local" section of development docs to be more accurate so people don't run into the same snags I did. it wasn't immediately clear that `.dev/` needed to be populated with the Core and Periphery configs and the DB needed to be run separately.

this is quite a bit of manual steps to get started and also to start every session (4 manual commands). i'd be happy to help with the DX here, but I don't want to move too many things at once, so just documenting the status quo for now.


mbecker20 (approved):
This ferretdb is old, will change to mongo

PR #1243, opened 2026-03-12
improve delete button styling

sloria:
the filled red delete buttons have too much visual weight, especially given that they all have a confirmation step after. this changes the pattern to: subtle styling (e.g. outline button) before first click -> filled red for confirm.

The Resource page headers are where this is most obvious.

Before:

header-before

after:

header-after

Similarly, the 2-click standalone buttons.

Before:

standalone-before

After:

standalone-after

Modals had the inverse logic.

Before:

modal-before

After:

modal-after

### DIFF
--- a/ui/src/components/swarm/remove.tsx
+++ b/ui/src/components/swarm/remove.tsx
@@ -35,7 +35,7 @@ export default function RemoveSwarmResource({
   return (
     <ConfirmModalWithDisable
       confirmText={resourceName ?? resourceId}
-      targetProps={{ variant: "filled", color: "red" }}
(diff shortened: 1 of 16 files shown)


sloria:
@mbecker20 did you mean to close this, or was this auto-closed due to the 2.0.0 merge?

mbecker20:
It was auto-closed, I also noted this and applied the same idea in a slightly different way in v2 (added `confirmProps`, kept `default` variant). Feel free to open another PR to main if you feel it still could be improved

sloria:
ah i see that v2 stable has changed the colors of the delete buttons. looks much better! thank you

PR #1265, opened 2026-03-24
let user specify if they want command to be run in shell-mode

ericls:
This is a follow up to https://github.com/moghtech/komodo/pull/1217 in v2.

In v2, things are easier because there's an explicit shell-mode. This makes this change easier to implement and won't cause backwards incompitable changes for existing configs.

Makes something like the following possible:

Screenshot 2026-03-24 at 19 30 41

Without this option, scripts like this would fail for reason like:
```
bash: This: command not found
```

Would improve #808

closes #1217

### DIFF
--- a/bin/periphery/src/api/build/mod.rs
+++ b/bin/periphery/src/api/build/mod.rs
@@ -271,7 +271,7 @@ impl Resolve<crate::api::Args> for build::Build {
         "Pre Build",
         pre_build_path.as_path(),
         &pre_build.command,
-        KomodoCommandMode::Multiline,
+        if pre_build.shell_mode { KomodoCommandMode::Shell } else { KomodoCommandMode::Multiline },
         &replacers,
       )
       .instrument(span)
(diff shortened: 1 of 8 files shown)


mbecker20:
Thanks will review

ericls:
> Thanks will review

Thanks for Komodo.
V2 has been fantastic!

FDHoho007:
I also need this because I started deploying config files via pre deploy actions. Currently I am basically force to squash everything into one line which makes it ugly to read or check the diff during resource sync.

Thanks for implementing @ericls.

mbecker20:
The formatting isn't standard, you can run `cargo fmt` in the repo and commit

ericls:
> The formatting isn't standard, you can run `cargo fmt` in the repo and commit

Ran `cargo fmt`, hope it's compliant now. Otherwise I may need to rebase first.

ericls:
hey @mbecker20 , let me know if this can be merged or any changes are needed. Happt to chat about any thoughts you may have!

mbecker20:
It looks good to me, thanks @ericls ! Will merge into 2.2.0 to include in next release.

PR #1266, opened 2026-03-25
Prevent horizontal page scroll on long log lines

elemated:
Small update to add white-space: pre wrap to a few places where logs are converted to HTML. In cases where logs printed a long single line, the line wasn't breaking. Horizontal scrolling the entire page was required to see logs, which was not a pleasant user experience.

**action/last-run.tsx**

Before (image 1: haven't scrolled yet, image 2: scrolled all the way to the right)
Screenshot 2026-03-24 at 5 53 11 PM
Screenshot 2026-03-24 at 5 53 15 PM

After:
Screenshot 2026-03-24 at 4 10 12 PM

Was not able to test changes for build/info.tsx as I don't use that feature, but have attached screenshots for the **updates/details.tsx** changes below as well. I believe they should all behave similarly.

Before (image 1: haven't scrolled yet, image 2: scrolled all the way to the right)
Screenshot 2026-03-24 at 5 49 51 PM
Screenshot 2026-03-24 at 5 49 55 PM

After:
Screenshot 2026-03-24 at 5 40 16 PM

### DIFF
--- a/ui/src/components/updates/details.tsx
+++ b/ui/src/components/updates/details.tsx
@@ -200,7 +200,7 @@ export function UpdateDetailsContent({ id }: { id: string }) {
                   dangerouslySetInnerHTML={{
                     __html: updateLogToHtml(log.stdout),
                   }}
-                  style={{ overflowY: "auto" }}
+                  style={{ overflowY: "auto", whiteSpace: "pre-wrap" }}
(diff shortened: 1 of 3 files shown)


elemated:
will re-open with additional UI fix

PR #1279, opened 2026-03-27
Allow excluding services from global auto update 

mateuszziolkowski:
Problem:
We have our Global Auto Update set to run every 3 minutes, which depletes docker hub limits really fast. Only one service in our stack needs to be constantly checked for updates (and it's on ghcr.io, so no rate limit problem). 
One of the solutions was to set specific image digests (@sha256:...) for all services using images from docker hub in compose.yaml, but then when you want to update - you have to manually check for digests and replace them in compose.yaml.

Solution:
This PR adds new option, which allows you to exclude services from Global Auto Updates checks (manual checks still look for new digest).

When poll for updates disabled:
obraz

When poll for updates enabled:
obraz

### DIFF
--- a/ui/src/resources/stack/config/index.tsx
+++ b/ui/src/resources/stack/config/index.tsx
@@ -95,6 +95,8 @@ export default function StackConfig({
   const disabled = global_disabled || !canWrite;
 
   const runBuild = update.run_build ?? config.run_build;
+  const poll_for_updates = update.poll_for_updates ?? config.poll_for_updates;
+
   const mode = getStackMode(update, config);
 
   const gitProvider = update.git_provider ?? config.git_provider;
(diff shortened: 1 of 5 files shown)


mbecker20 (changes requested):

ui/src/resources/stack/config/index.tsx
```diff
@@ -95,6 +95,8 @@ export default function StackConfig({
   const disabled = global_disabled || !canWrite;
 
   const runBuild = update.run_build ?? config.run_build;
+  const poll_for_updates = update.poll_for_updates ?? config.poll_for_updates;
```
mbecker20, line 98: `poll_for_updates` also implicitly true if `auto_update` is true. You don't need to set both of them to true. This check also needs to check auto_update, but I can fix

mbecker20 (approved):
LGMT, I will make the small fix after merge

PR #1326, opened 2026-04-04
Use `build.get_image_tags` for deployments from builds

### LINKED ISSUE #1325: Deploy tries to find image tagged with version while build has "Include Version Tags" disabled
When trying to deploy a build, it fails with this error:

Image

Here's how deploy source is set up:

Image

And here's the build's image tags settings:

Image

Probably should use `:latest` when "Include Version Tags" is toggled off?

4ndv:
- Reuse `build.get_image_tags()` for building image tag for deployments
- Reorder stuff in `get_image_tags` for preserving current deployment behavior (to prefer full version tags). It's currently only used for running builds, and builder doesn't really care about the order
- Fallback to `:latest` for both builds and deployments if no other tags were added to not fail when all of the tagging options are disabled

Tried my best to test all the combinations, but I'm not very familiar with the codebase and may miss some things

Fixes #1325

### DIFF
--- a/bin/core/src/api/execute/deployment.rs
+++ b/bin/core/src/api/execute/deployment.rs
@@ -123,26 +123,24 @@ impl Resolve<ExecuteArgs> for Deploy {
       DeploymentImage::Build { build_id, version } => {
         let build = resource::get::<Build>(build_id).await?;
         let image_names = build.get_image_names();
-        let image_name = image_names
+        let tags = build.get_image_tags(
+          &image_names,
+          build.info.built_hash.as_deref(),
+          &vec![], // Unused for now
+        );
+        let image_name = tags
           .first()
           .context("No image name could be created")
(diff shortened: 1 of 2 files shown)


mbecker20:
Thanks for catching and fixing, will review shortly

bin/core/src/api/execute/deployment.rs
```diff
@@ -308,26 +306,23 @@ pub async fn pull_deployment_inner(
   server: &Server,
 ) -> anyhow::Result<Log> {
   let (image, account, token) = match deployment.config.image {
-    DeploymentImage::Build { build_id, version } => {
+    DeploymentImage::Build {
+      build_id,
+      version: _,
```
mbecker20, line 311: version should not be unused here, please review

mbecker20:
Fixed slightly differently in 2.3.0

PR #1337, opened 2026-04-07
Add support to generate resource json schema

bytedream:
Adds support to generate a json schema for resource definitions. This allows for suggestions in editors:
image

The schema is generated via the `schemars` crate, it provides a derive macro which is placed above each struct that is relevant to resource definitions (locked behind the `schemars` feature). It can be generated by a subcommand of the new created `xtask` crate: e.g. `cargo xtask generate resource-json-schema --pretty --stdout`.

Although `schemars` is serde compatible in most parts, it's currently unable to create schema definitions for aliases. I've manually renamed fields where an alias is more understandable than the rust field name (via field attributes, the renames are exclusive to `schemars`). I've also looked if `typeshare` can generate a json schema, but it doesn't seem so.

If the PR gets accepted, it would be nice if the schema could be hosted on the komodo website, so users can directly include the url in their config files / LSPs and don't have to manually download / generate it.<br>It might also be possible to generate parts of the resource docs on the komodo website by the json schema, but I haven't looked further into it.

---

Relies on https://github.com/mbecker20/partial_derive2/pull/1 to compile (hence the changed `partial_derive2` dependency, the rev points to the PR head) and https://github.com/mbecker20/partial_derive2/pull/2 to include all comments.


mbecker20:
Looks nice, will review shortly

mbecker20 (approved):
LGTM, thanks! Including in 2.2.0

mbecker20:
BTW, anything in the `ui/public` folder is hosted by every Komodo Core instance. The build step can include schema gen and copy to this folder. Then you can load from your own Komodo (or https://demo.komo.do).

mbecker20:
Will also try loading the schema into the web UI monaco editor, then in UI edit of resource toml may have the suggestions

PR #1365, opened 2026-04-18
Minor: Update periphery.config.toml

andriibratanin:
Make docs on ROOT_DIRECTORY more clear

This change relates to line:
> Each specific sub-directory (like ./stacks) can be overridden below.

The change includes `./builds` directory to the list as it also can be overridden. Thus, a reader will have better understanding of the resulting root directory structure


mbecker20 (approved):
LGTM

PR #1366, opened 2026-04-18
Move from yarn link to yarn workspace

bytedream:
Replace `yarn link` logic with yarn workspaces. This significantly improves the development and building experience and is more straightforward imo.


mbecker20:
Will review shortly

Issue #1377, opened 2026-04-21
[bug] Registry account password/token with ' (apos) char


marchrius:
When adding a custom Sonatype Nexus 3 registry with ' char (apos) this is the error we have when pull images

>ERROR: Failed to login to image registry
>
>TRACE:
>	1: Domain: 'registry.example.com' | Account: 'nexus'
>	2: Registry registry.example.com login error
>	3: /bin/bash: -c: row 1: EOF unexpected during the search for "'"
>	4: End of trace

The password is something like **.Aa'xxxx**

A workaround is to not use an apostrophe in the password

marchrius:
Here just to add that the password cannot have chars like $ (dollar sign)

PR #1395, opened 2026-04-27
[bug fix] pass docker login token via stdin

marchrius:
Ref. #1377 

Avoid interpolating registry tokens into a shell command during docker login. This preserves passwords containing shell-special characters such as quotes or dollar signs.

## Summary

  Fix Docker registry login handling in Periphery by passing the registry token directly to `docker login --password-stdin` instead of interpolating it into a shell command.

  ## Why

  Registry passwords or tokens containing shell-special characters like `'` or `$` could break login because they were evaluated by the shell before reaching Docker.

  ## Changes

  - Replace `echo {token} | docker login ...` with direct `tokio::process::Command` execution.
## Changes

  - Replace `echo {token} | docker login ...` with direct `tokio::process::Command` execution.
  - Write the registry token to the Docker process stdin.
  - Preserve existing Docker login error handling behavior.

  ## Verification

  - `cargo fmt --check`
  - `git diff --check`
  - `cargo check -p komodo_periphery`
  - VSCode `Init` task equivalent completed

### DIFF
--- a/bin/core/src/network.rs
+++ b/bin/core/src/network.rs
@@ -259,9 +259,10 @@ async fn set_default_gateway(
     ));
   }
 
-  // Remove existing default routes
-  let remove_default = Command::new("sh")
-    .args(["-c", "ip route del default 2>/dev/null || true"])
+  // Remove existing default routes. Failure is ignored because the route may
+  // not exist yet.
+  let remove_default = Command::new("ip")
+    .args(["route", "del", "default"])
     .output()
     .await;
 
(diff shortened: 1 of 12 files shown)


bytedream:
There are multiple occurrences where this can happen, `run_shell_command` is also internally called by `run_komodo_shell_command`, which is used in different places with string interpolation. Rather than fixing this one specific command, the whole shell execution model should move away from `sh -c` IMO, as it's inherently insecure (breaking the command as reported in the issue is only the best case, this might also lead to remote-code execution).

marchrius:
I tried to address this by centralizing safe command execution in lib/command.
  Internal commands now pass the program and arguments separately instead of interpolating through a shell.
  I migrated the riskier paths: Docker login, log search, secret/config creation, stop/remove, and internal Docker/Compose commands.
  Shell execution remains only where intentional: user scripts, configured wrappers, and cases requiring explicit shell behavior.
  This will definitely need review from a maintainer who knows the app better than I do; I’m trying to help with the support of AI.

mbecker20:
v2.3.0 and after passes the token via stdin

PR #1463, opened 2026-06-04
Allow to specify KOMODO_HOST as URL with subpath

### LINKED ISSUE #353: [Feature Request] Set a custom base url
I want to be able to put komodo on a custom subpath behind a reverse proxy, for example:

https://example.com/komodo

This would require all URLs be updated to prefix with the subpath, so all subsequent request from the browser are made to the correct URL.

I see the option for setting a custom `webhook_base_url` for the webhook, but nothing to change it for the whole application.

lafriks:
Fixes #353

I'm not rust developer so rust code was completely made by LLM but tested and it works. Anyway feel free to remake it but hope it could be used as inspiration on how this can be made to work to specify env `KOMODO_HOST=https://my.domain/komodo/`


lafriks:
@mbecker20 anything I can do/improve to get this merged?

PR #1486, opened 2026-06-18
Send keepalive pings on user websockets to survive idle proxy timeouts

Dougley:
Cloudflare closes TCP sockets that go quiet, and our two user-facing sockets `/ws/update` and `/ws/terminal`, send nothing while idle. The socket sits there until the proxy reaps it and the UI falls back to its reconnect loop. 

This PR makes the server send a WebSocket Ping every 30s on both endpoints. 

(Core<->Periphery connections already ping every 5s. This only touches the two browser-facing endpoints.)

## Files
  - `bin/core/src/api/ws/mod.rs` - shared `WS_KEEP_ALIVE_INTERVAL` (30s)
  - `bin/core/src/api/ws/update.rs` - ping in the update send loop
  - `bin/core/src/api/ws/terminal.rs` - ping in the terminal forward loop

### DIFF
--- a/bin/core/src/api/ws/mod.rs
+++ b/bin/core/src/api/ws/mod.rs
@@ -24,6 +24,9 @@ use crate::{
 mod terminal;
 mod update;
 
+pub(super) const WS_KEEP_ALIVE_INTERVAL: std::time::Duration =
+  std::time::Duration::from_secs(30);
+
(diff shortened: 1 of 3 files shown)


mbecker20 (approved):
Looks good to me, thanks

PR #1489, opened 2026-06-20
feat: break out cache/buffers and ZFS ARC from used memory

ChanningHe:
On hosts with a big page cache or ZFS ARC, used memory looks almost full and the memory alerts keep firing, even though that memory is reclaimable. ZFS is the worst case since `MemAvailable` doesn't count the ARC. There are also a few open issues about this.

Periphery now reports memory broken out into used, cache/buffers, ZFS ARC and swap, and keeps reclaimable memory out of "used" so the alerts track real pressure. 

Notes for review:
- `bin/periphery/src/stats.rs` was moved to `stats/mod.rs` to make room for the new `stats/mem.rs` (the /proc parsing and calculation).
- `historical.tsx` has the largest diff — the memory chart went from a single line to a stacked composition.

Screenshot:
RAM Usage
Memory


mbecker20 (approved):
Very nice, thanks

Issue #1392, opened 2026-04-27
Periphery Pending keepalive defeats Core's RPC timeout: hung subprocess on one host can wedge a deploy and freeze fleet-wide Stats writes


vvv850:
A `docker compose ls` child on one of our v2.1.2 periphery hosts hung in `futex_wait_queue` (looked like a leftover from a `context canceled` during an image manifest read). Subprocess hangs happen. What surprised us was the blast radius:

- A `DeployStack` against that host hung for 104 minutes with the UI stuck on "Deploying…". Only restarting Core ended it. `DestroyStack` as a workaround was rejected as `Resource is busy`.
- `Stats` writes for **all 6 servers in the fleet** stopped at the exact same instant the subprocess hung, and resumed the moment we killed the orphan PID. ~46 hours of frozen monitoring.
- No alert. Periphery's websocket login flag stayed green throughout.
- After we manually restarted Core to break the hung deploy, all 6 peripheries re-logged in over websocket within ~1s, but one stayed shown as "down" in the UI until we killed the orphan PID. During that window Core spammed `WARN Failed to forward Response message | No response channel found at <uuid>` every 5s for one channel_id, which is periphery's keepalive still firing for the wedged pre-restart request after Core's matching channel was gone. So restarting Core is not a real recovery here: the wedged tokio task on the periphery outlives Core's lifecycle.

The chain in sourc:

- [`lib/command/src/lib.rs:160`](https://github.com/moghtech/komodo/blob/v2.1.2/lib/command/src/lib.rs#L160): `cmd.output().await` has no timeout, so a hung child pins the future. `kill_on_drop(true)` is set but never fires because nothing drops the future.
- [`bin/periphery/src/connection/mod.rs:182`](https://github.com/moghtech/komodo/blob/v2.1.2/bin/periphery/src/connection/mod.rs#L182): request handler runs in a `tokio::select!` against an infinite `ping_in_progress` loop sending `Pending` every 5s. The keepalive stops only when `resolve_response` finishes, which it never will.
- [`bin/core/src/periphery/mod.rs:138`](https://github.com/moghtech/komodo/blob/v2.1.2/bin/core/src/periphery/mod.rs#L138): Core loops on `recv().with_timeout(10s)` and `continue;`s on every `Pending`. Keepalives every 5s mean the effective RPC timeout is infinite.
[...]

Issue #1289, opened 2026-03-28
[Bug] "execute_container_terminal" randomly fails to execute


bastienlc:
## Setting

*komodo-core and komodo-periphery running 2.0.0 docker image*

Assume a debian:bookworm container is running on "my-server":
```bash
docker run --name debug -it debian:bookworm /bin/sh
```
I'm trying to execute a command on this container using an Action, specifically the `execute_container_terminal` method of the typescript client. Thus from the UI I create an action with this content:
```ts
let shell = "sh"; // sh or bash
let server = "my-server";
let container = "debug";
let command = "date";

await komodo.execute_container_terminal(
  {
    server,
    container,
    command,
    init: { command: shell, recreate: "Always" },
  },
  {
    onLine: (line) => {
      console.log(line);
    },
    onFinish: (code) => {
      console.log(`Command finished with exit code ${code}`);
    },
  },
);
```

## Expected behavior

Using `bash` or `sh`, executing the action succeeds and the following output is produced:
```bash

Sat Mar 28 20:31:22 UTC 2026

Command finished with exit code 0
🦎 Action completed successfully 🦎
```

## Actual behavior

Sometimes the command is executed incorrectly, resulting in the following output:
```bash

'; date; rc=$?; printf '
Command finished with exit code %d
🦎 Action completed successfully 🦎
```

I have not been able to pinpoint exactly what causes this behavior to occur, currently it seems random.

## Initial analysis

While I'm not familiar with the komodo source code, the issue seems to stem from this snippet:
https://github.com/moghtech/komodo/blob/0f9679455911ce9f078e1d3654437ce4cda84eda/bin/periphery/src/api/terminal.rs#L466-L468

## Edits

- Updated behavior from "succeeds with bash but fails with sh" to "randomly fails"

martkaczmarek:
Happening to me too. 
Komodo periphery  + core both on Docker, 2.1.2

In my case, even multiple executes across a singe action could change behavior (this is output of a single action):

```
execute_stack_exec bash:

Tue Apr 14 19:07:41 UTC 2026       <--"date" command executed properly

Exit code: 0
execute_stack_exec sh:

'; date; rc=$?; printf '       <--"date" command printed as is
Exit code: %d
execute_stack_service_terminal bash:

'; date; rc=$?; printf '
Finished. Code:  %d
execute_stack_service_terminal sh:

'; date; rc=$?; printf '
Finished. Code:  %d
🦎 Action completed successfully 🦎
```

[...]

Issue #868, opened 2025-09-21
[Bug] Issue With Variable Interpolation


meerumschlungen:
Hi folks,
thanks for your great work! I've got this Stack configuration deployed from a git repo. The .env file is written on pulling the repo.
<details>
<summary>compose.yaml (exactly)</summary>

```yaml
---
name: semaphore
services:
  postgres:
    image: docker.io/library/postgres:17.6@sha256:feff5b24fedd610975a1f5e743c51a4b360437f4dc3a11acf740dcd708f413f6
    volumes:
      - db_data:/var/lib/postgresql/data
    environment:
      POSTGRES_USER: ${SEMAPHORE_DB_USER}
      POSTGRES_PASSWORD: ${SEMAPHORE_DB_PASS}
      POSTGRES_DB: ${SEMAPHORE_DB_NAME}
    networks:
      - default
    restart: unless-stopped
  semaphore:
    image: semaphoreui/semaphore:v2.16.31@sha256:7c9617ecd6233a019c85f52b122108c1113458c3cf91554145f3c56d4dbc25b3
    depends_on:
      - postgres
    env_file: .env
    networks:
      - default
      - traefik_services
    restart: unless-stopped
    labels:
      - traefik.enable=true
      - traefik.http.routers.semaphore.entrypoints=websecure
      - traefik.http.routers.semaphore.rule=Host(`${FQDN}`)
networks:
  default: {}
  traefik_services:
    external: true
volumes:
  db_data:

```

</details>
<details>
<summary>Environment generation configuration when pulling the repo (here the variable substitution happens)</summary>

```sh
FQDN=semaphore.redacted.com

SEMAPHORE_DB_DIALECT=postgres
SEMAPHORE_DB_HOST=postgres
SEMAPHORE_DB_NAME=semaphore
SEMAPHORE_DB_USER=[[SEMAPHORE__SEMAPHORE_DB_USER]]
SEMAPHORE_DB_PASS=[[SEMAPHORE__SEMAPHORE_DB_PASS]]
SEMAPHORE_ADMIN=[[SEMAPHORE__SEMAPHORE_ADMIN]]
SEMAPHORE_ADMIN_PASSWORD=[[SEMAPHORE__SEMAPHORE_ADMIN_PASSWORD]]
SEMAPHORE_ADMIN_NAME=[[SEMAPHORE__SEMAPHORE_ADMIN]]
[...]

PR #1520, opened 2026-07-10
2.3.0

mbecker20:
# Changelog

## Pagination
- Add pagination for resource and container list APIs, improved term based search and server-side sorting. Use Komodo with lots of resources without having issues due to over-querying.
    - API users can still disable pagination by passing `limit: 0` in the request body json. The default limit is 50 resources per page.
- **UI / Containers**: Improved page responsiveness and container search with arbitrarily many containers using pagination.

Screenshot 2026-07-11 at 1 28 37 AM

## RAM / Monitoring
- Break down memory usage into used / cache, supporting ZFS, by @ChanningHe.

Memory

## Cancellation

- **Build**: All builds now support mid-build cancellation by terminating the running build command.
    - Build webhooks now cancel any running build before re-triggering, better handling cases with back-to-back git pushes.
- **Action / Procedure**: Support mid-run cancellation.
    - Actions cancel immediately by killing the deno process executing the script.
    - Procedures cancel after the current parallel stage completes (before moving onto the next stage)

Screenshot 2026-07-11 at 1 37 09 AM

## Stats page
- Top level stats viewer including multi-server charts

Screenshot 2026-07-24 at 8 03 30 PM

## Misc.

- **Procedure**: Batch executions now support matching by tags.
    - Specify a pattern of `*` to filter only by tags.
- **Builder**: Server type builders can now attach multiple servers and simultaneous builds will be distributed between them.
- **Deployment**: Support configuring custom name for container / swarm service
    - The container / service name still defaults to deployment name
- **Stack**: Allow viewing service list even when down, enabling users to deploy single service without the whole Stack.
- **Auth**: Improve Google auth integration with id token validation by @MP-Tool 
- **UI**: Add omni-search 'clear on close' option.
- **UI**: Find Stack services from the omni-search.
- **UI**: Fix permissions table performance.
- **Commands**: Add command-specific timeouts on shell commands. Fixes issues like #1392.
- **Reporting**: Add **opt-in** reporting about Komodo usage.
- **Action**: Fix `execute_terminal` output capturing using `sh` by @mvanhorn, fixing #1289 
- **Build / Deployment**: Fix build -> deployment flow when build doesn't push semver tags.
- **Stack**: Fix compose config sanitization ordering by @mvanhorn and @mbecker20 re #868 
- **Alerter / TOML**: Fix resource lists export resources IDs instead of names.
- **Git**: Fix giving clear log when `git` isn't installed, previously the message did not clearly identify the issue in logs.
- **UI**: Fix resource description responsive overflow.
- **UI / Websocket**: Send keepalive pings to help websocket maintain connection by @Dougley.
- **UI / Deployment**: Fix deployment tabs disappearing sometimes when in swarm mode.


sloria:
this looks great! one nitpick: in the same vein as https://github.com/moghtech/komodo/pull/1272 , the "Cancel" buttons carry a lot of visual weight as implemented and should probably be non-red. ofc no need to block release over it though. thanks for all the awesome work

Issue #1523, opened 2026-07-13
setup-periphery.py missing option to pass core_public_keys


amosshapira:
When trying to run a periphery which was setup using [setup-periphery.py](https://github.com/moghtech/komodo/blob/main/scripts/setup-periphery.py), it fails with the following log:

```
Periphery failed to validate Core public key:
Ensure public key matches one of the 'core_public_keys' in periphery config
(PERIPHERY_CORE_PUBLIC_KEYS): MCowBQYDK2VuAyEA6JKHQw6I0l73MhUT2jML2sNY7j8aL4zR8U05Rg3uwVE= is invalid
```

And indeed, `setup-periphery.py` doesn't provide a way to set the value of `core_public_keys`.

Suggestion: add a --core-public-keys flag to `setup-periphery.py` which can accept one or more values to put in the generated file.

PR #1524, opened 2026-07-13
setup-periphery.py: add "--core-public-keys" command line argument

amosshapira:
Addressed issue https://github.com/moghtech/komodo/issues/1523

Tested this on my setup and the periphery server was registered correctly by the Core.


mbecker20 (approved):
LGTM

PR #1528, opened 2026-07-17
fix: log webhook branch-mismatch skips at debug instead of warn

cplieger:
## Problem

Git providers send push webhooks for **every** branch, and Komodo filters them server side against the branch configured on the resource — exactly as the docs describe (Automate → Webhooks → Branch Filtering). But the skip is implemented as an ordinary error, so it surfaces through the listener's generic failure log:

```
WARN ProcedureWebhook{id="..."}: Failed at running webhook for procedure ... | target branch: main | request branch does not match expected
```

Two problems with that:

- A routine, by-design skip is indistinguishable from a genuine failure (payload parse error, execution error) — same message shape, same WARN level, same log site.
- Any repo with active bot branches makes it noisy. A repo receiving Renovate PRs produces a WARN for every `renovate/*` branch push; on my instance that is **hundreds of warns per day** across two listeners, drowning the warns that matter.

It is also inconsistent with the other by-design skip: `webhook_enabled = false` returns `Ok(())` silently.

## Change

Rename `ExtractBranch::verify_branch` → `branch_matches`, returning `anyhow::Result<bool>`:

- **Branch mismatch** → `Ok(false)`, logged once at `debug!` (`Ignoring webhook | push to branch '...' does not match expected branch '...'`), and the handler returns `Ok(())` — a quiet skip, consistent with the `webhook_enabled` skip. The resource id still lands on the event via the existing `*Webhook{id}` span.
- **Branch extraction failure** (e.g. a ref-less `release`/`workflow_run` payload delivered to a named-branch listener) → still `Err`, still reaching the `Failed at running webhook` WARN. That case is a real misconfiguration signal and keeps its visibility.

The Action handler keeps its inline comparison (it extracts the branch anyway for `WEBHOOK_BRANCH`) with the same debug-and-skip behavior.

No behavior change beyond the log level: executions trigger (and don't trigger) exactly as before.

## Testing

- `cargo check -p komodo_core` clean on this branch.
- `cargo fmt --all -- --check` clean.
- `cargo test`: the 2 failures in `lib/command` (`cancel_kills_process_group`, `timeout_kills_process_group`) are pre-existing on an unpatched checkout of `2.3.0` in my sandbox and unrelated to this change (listener code has no tests).
- Live-traffic sanity check of the current behavior (v2.2.0): confirmed each `renovate/*` push produces the WARN above while a matching `main` push triggers the procedure.

### DIFF
--- a/bin/core/src/api/listener/mod.rs
+++ b/bin/core/src/api/listener/mod.rs
@@ -1,6 +1,5 @@
 use std::sync::Arc;
 
-use anyhow::anyhow;
 use axum::{Router, http::HeaderMap};
 use komodo_client::entities::resource::Resource;
 use mogh_cache::CloneCache;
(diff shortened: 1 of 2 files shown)


mbecker20 (approved):
Makes sense, thanks

PR #1533, opened 2026-07-21
fix(core): remove double offset in GetHistoricalServerStats pagination

### LINKED ISSUE #1532: GetHistoricalServerStats returns empty results after the first page
## Summary

`GetHistoricalServerStats` paginates by applying the `page` offset **twice**: once when building `ts_vec` (shifting the time window back by `granularity * STATS_PER_PAGE * page`), and again via `.skip(page * STATS_PER_PAGE)` on the MongoDB query. Because `$in: ts_vec` already restricts the result to at most `STATS_PER_PAGE` documents for that page, the subsequent `.skip(page * STATS_PER_PAGE)` discards every matched document when `page >= 1`, so every page after the first always returns an empty `stats` array.

## Location

`bin/core/src/api/read/server.rs`, `GetHistoricalServerStats::resolve` (around lines 317–368).

```rust
const STATS_PER_PAGE: i64 = 200;

impl Resolve<ReadArgs> for GetHistoricalServerStats {
  async fn resolve(self, ReadArgs { user }: &ReadArgs) -> mogh_error::Result<GetHistoricalServerStatsResponse> {
    let GetHistoricalServerStats { server, granularity, page } = self;
    // ... permission check ...
    let granularity = get_timelength_in_ms(granularity.to_string().parse().unwrap()) as i64;
    let mut ts_vec = Vec::<i64>::new();
    let curr_ts = unix_timestamp_ms() as i64;
    let mut curr_ts = curr_ts
      - curr_ts % granularity
      - granularity * STATS_PER_PAGE * page as i64;   // <-- offset #1: time window shifted by page
    for _ in 0..STATS_PER_PAGE {
      ts_vec.push(curr_ts);
      curr_ts -= granularity;
    }

    let stats = find_collect(
      &db_client().stats,
      doc! { "sid": server.id, "ts": { "$in": ts_vec } },   // <-- already pinned to this page's 200 slots
      FindOptions::builder()
        .sort(doc! { "ts": -1 })
        .skip(page as u64 * STATS_PER_PAGE as u64)           // <-- offset #2: skips all of the above
        .limit(STATS_PER_PAGE)
        .build(),
    )
    .await
    .context("failed to pull stats from db")?;

    let next_page = if stats.len() == STATS_PER_PAGE as usize {
      Some(page + 1)
    } else {
      None
    };
    // ...
  }
}
```

## Why page >= 1 is always empty

1. `ts_vec` is constructed as the 200 granularity-slots ending at `now - now%g - g*200*page`. For `page = 1` with `granularity = 1-min`, that is the 200 minute-slots from `now - 400 min` to `now - 201 min`.
2. The filter `{ ts: { $in: ts_vec } }` matches **only** documents whose `ts` is in those 200 slots — at most 200 documents.
3. `.skip(page * 200)` then skips `page * 200` documents from that already-bounded set. For `page = 1`, it skips 200, which is the entire matched set → `stats` is empty.
4. `next_page` becomes `None` (because `stats.len() != STATS_PER_PAGE`), so callers stop and silently lose all data older than the first page.

`page = 0` is unaffected because `.skip(0)` is a no-op.

## Reproduction

Tested against a Komodo Core instance on `main` (file tree fetched from `moghtech/komodo` main branch).

Server `jw-test` has stats recorded from `2026-07-20 21:00` through `2026-07-21 15:00` (verified via `5-min` and `1-hr` granularity, both single-page).

### `granularity = "1-min"`

```
page 0: len=200  next_page=1  oldest=2026-07-21 11:47  newest=2026-07-21 15:06
page 1: len=0    next_page=None   (empty)
```

`page = 1`'s `ts_vec` covers `2026-07-21 08:28 ~ 11:46`. Data **does** exist at those minute boundaries — a `5-min` query returns 33 points in that window, all with `ts % 60000 == 0` (e.g. `ts = 1784605500000` = `2026-07-21 11:45:00`). The `$in` filter matches them; `.skip(200)` then discards them.

### `granularity = "5-min"` (single page, not affected by the bug)

```
page 0: len=189  next_page=None  oldest=2026-07-20 22:30  newest=2026-07-21 15:05
```

### `granularity = "1-hr"` (single page, not affected)

```
page 0: len=18   next_page=None  oldest=2026-07-20 21:00  newest=2026-07-21 15:00
```

## Expected behavior

`page = n` should return the `n`-th 200-slot window of historical stats counting back from `now`, and `next_page` should reliably indicate whether an older window has data.

## Suggested fix

The `ts_vec` already encodes the page offset, so `.skip()` should be removed (the `$in` filter is the correct paging mechanism here):

```rust
let stats = find_collect(
  &db_client().stats,
  doc! { "sid": server.id, "ts": { "$in": ts_vec } },
  FindOptions::builder()
    .sort(doc! { "ts": -1 })
    // no .skip() — ts_vec already selects the page's time window
    .limit(STATS_PER_PAGE)
    .build(),
)
.await
.context("failed to pull stats from db")?;
```

Alternatively, if `.skip()` is kept for some reason, `ts_vec` must not be shifted by `page` (build it from `now` every time and let `skip/limit` do the paging). The current code applies both, which is the bug.

## Environment

- Komodo Core: `main` branch (`bin/core/src/api/read/server.rs` as of report date)
- `STATS_PER_PAGE = 200`, default `monitoring_interval = 15-sec`, default `keep_stats_for_days = 14`
- Reproduced with `granularity` values `1-min`, `5-min`, `1-hr` against a live instance

## Impact

Any consumer that walks `next_page` to fetch history longer than `STATS_PER_PAGE * granularity` (e.g. 200 minutes for `1-min`, ~3.3 hours) silently receives only the most recent 200 points and no error. Weekly/monthly reports built on this API are incomplete with no signal of data loss.

CaiJingLong:
## Related Issue
Closes #1532

## Problem
`GetHistoricalServerStats` applies the `page` offset twice: once when building `ts_vec` (shifting the time window back by `granularity * STATS_PER_PAGE * page`), and again via `.skip(page * STATS_PER_PAGE)` on the MongoDB query. Since the `{ ts: { $in: ts_vec } }` filter already restricts the result to at most `STATS_PER_PAGE` documents for that page, the subsequent `.skip(page * STATS_PER_PAGE)` discards every matched document when `page >= 1`, so every page after the first always returns an empty `stats` array and `next_page` becomes `None`, silently losing all older historical data.

## Change
- `bin/core/src/api/read/server.rs`, `GetHistoricalServerStats::resolve`: removed the `.skip(page * STATS_PER_PAGE)` call from `FindOptions::builder()`. `ts_vec` already encodes the page offset via the time window shift, so the `$in` filter is the correct paging mechanism here — no additional skip is needed.

## Verification
- `cargo check -p komodo_core` passes (no new warnings).
- `cargo fmt -p komodo_core --check` passes.
- Cross-checked against the issue reproduction: `page = 0` behavior is unchanged (`.skip(0)` was already a no-op); `page >= 1` is no longer wiped out by the second offset and now returns the documents matching the `ts_vec` time window, with `next_page` correctly indicating whether an older window has data.
- `bin/core` has no existing unit-test infrastructure and the function depends on a live MongoDB runtime, so no new tests were added; the fix can be verified manually via the issue's `1-min` multi-page reproduction steps.

## Risk
- Change is a single-line removal; does not affect the `page = 0` path. The `ts_vec` construction and `next_page` logic are unchanged. No schema migration, config change, or breaking API change.
- Note: `next_page` is `Some(page + 1)` only when the page returns exactly `STATS_PER_PAGE` records. If a page returns fewer than 200 due to missing data while an older window still has data, `next_page` will be `None` — this is pre-existing behavior, unchanged by this fix, which only removes the double offset that caused the spurious empty results.

## Conventions
- The repository has no explicit `CONTRIBUTING.md` / `CLAUDE.md` / `AGENTS.md` / PR template, so common conventions (Conventional Commits + default PR structure) are used.

### DIFF
--- a/bin/core/src/api/read/server.rs
+++ b/bin/core/src/api/read/server.rs
@@ -353,7 +353,6 @@ impl Resolve<ReadArgs> for GetHistoricalServerStats {
       },
       FindOptions::builder()
         .sort(doc! { "ts": -1 })
-        .skip(page as u64 * STATS_PER_PAGE as u64)
         .limit(STATS_PER_PAGE)
         .build(),
     )


mbecker20:
Fixed in 2.3.0

PR #1558, opened 2026-08-02
feat: :sparkles: S.M.A.R.T status report

GerardPolloRebozado:
This PR is a basic implementation of S.M.A.R.T status report by showing a badge in the list of drives
image
By having this status a new alert could also be implemented so if a device starts to fail it send alerts about it.
In order for this to work the docker compose needs to have access to the devices to check its health, this can be done by enabling [`SYS_RAWIO`](https://docs.docker.com/engine/containers/run/#:~:text=SYS%5FRAWIO,2%29%29%2E) or setting the container as privileged.
Also the device needs to be added in the devices property like this:
```yml
    devices:
      - /dev/nvme0n1
``` 
This is just an initial implementation on how the data could be gathered, I have not found another way to it with less privileges, other monitoring tools like [scrutiny](https://github.com/AnalogJ/scrutiny/blob/master/docker/example.hubspoke.docker-compose.yml) require same permissions. The current state of the PR only reports healthy or not and it does not trigger any alerts.
In the case the container doesn't have enough permissions the health will just show N/A

### DIFF
--- a/bin/core/src/alert/discord.rs
+++ b/bin/core/src/alert/discord.rs
@@ -117,12 +117,23 @@ pub async fn send_alert(
       path,
       used_gb,
       total_gb,
+      healthy,
+      temperature,
     } => {
       let region = fmt_region(region);
       let link = resource_link(ResourceTargetVariant::Server, id);
       let percentage = 100.0 * used_gb / total_gb;
+      let mut extra = String::new();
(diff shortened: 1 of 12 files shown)


GerardPolloRebozado:
I have also implemented showing both temperature and power on hours and alerts when there are high temperatures or a failed SMART test

PR #1599, opened 2026-08-20
fix: preserve tag filters during procedure batch expansion

### LINKED ISSUE #1596: Bug: BatchPullStack tag filters are ignored when executed inside a Procedure
### Environment

Komodo version: v2.3.2

### Description

When using BatchPullStack directly, the action supports both a stack matching pattern and a list of tags:

execution.type = "BatchPullStack"
execution.params.pattern = "*"
execution.params.tags = ["auto-pull"]

The expected behavior is that only stacks which Match the pattern, and Have the specified tag are selected.
However, when BatchPullStack is used inside a Procedure, the tags filter appears to be ignored completely.

For example, I have stacks tagged `auto-pull` which should be included in a scheduled pull, while locally built stacks are tagged `local-build` and intentionally do not have the auto-pull tag.

My Procedure configuration is:

[[procedure]]
name = "AutoPull"
config.webhook_enabled = false
config.schedule = "Run every day at 4:00 am"

[[procedure.config.stage]]
name = "Pull"
enabled = true
executions = [
  { execution.type = "BatchPullStack", execution.params.pattern = "*", execution.params.tags = ["auto-pull"], enabled = true }
]

Despite this, the Procedure attempts to execute PullStack against stacks which do not have the auto-pull tag.

Example sanitized error:

ERROR: Failed stage 'Pull' execution

TRACE:
    1: ERROR: Failed on PullStack(PullStack { stack: "example-local-stack", services: [] })
    2: ERROR: Execution not successful

The affected stack only has tags such as local-build & server-group and does not have auto-pull.

Expected behavior

Given:

execution.params.pattern = "*"
execution.params.tags = ["auto-pull"]

only stacks containing the auto-pull tag should be expanded and executed. Stacks without the tag should not receive a PullStack execution.

Actual behavior

All stacks matching the pattern appear to be expanded into individual PullStack executions regardless of execution.params.tags.

Possible cause

Looking at the v2.3.2 source, BatchPullStack itself correctly contains both:

pub pattern: String,
pub tags: Vec<String>,

and the normal batch execution path passes both the pattern and tag list into the resource query.

However, Procedure execution appears to expand batch operations before running them.

The relevant path appears to do roughly:

Execution::BatchPullStack(exec) => {
    extend_batch_exection::<BatchPullStack>(
        &exec.pattern,
        &mut executions,
    )
    .await?;
}

The exec.tags value is not passed to extend_batch_exection.

The expansion helper then calls:

list_full_for_user_using_pattern::<E::Resource>(
    pattern,
    Default::default(),
    ...
)

which means the resource query has no tag filter.

As a result, a Procedure containing:

pattern = "*"
tags = ["auto-pull"]

appears to effectively become:

pattern = "*"
tags = []

during Procedure batch expansion.

I also noticed the same batch expansion pattern appears to be used for other Procedure batch actions, so this may affect more than BatchPullStack.

Suggested fix

It would be useful for Procedure batch expansion to preserve the full filtering configuration rather than only the pattern.

For example, extend_batch_exection could accept the tag list:

extend_batch_exection::<BatchPullStack>(
    &exec.pattern,
    &exec.tags,
    &mut executions,
)

and construct the resource query using those tags instead of:

Default::default()

This would make Procedure behavior consistent with invoking the corresponding Batch API directly.

mvanhorn:
Update each Procedure batch match arm to pass the request's tags into the shared expansion helper along with its pattern. Make the helper build the same tag-aware `ResourceQuery` used by direct batch execution, loading the tag catalog only when a non-empty filter must be resolved and retaining the existing no-filter behavior for an empty list. Procedure stages expand batch executions into individual executions before running them. That expansion currently forwards each batch request's `pattern` but replaces its resource query with the default, so `tags` are discarded and stacks outside the requested tag set can be selected.

A Procedure `BatchPullStack` execution with pattern `*` and tag `auto-pull` expands only stacks carrying `auto-pull`, excluding a stack tagged only `local-build`; A Procedure batch execution with multiple requested tags preserves the default all-tags matching behavior instead of dropping or weakening the filter.

Closes #1596

### DIFF
--- a/bin/core/src/helpers/procedure.rs
+++ b/bin/core/src/helpers/procedure.rs
@@ -13,6 +13,7 @@ use komodo_client::{
     permission::PermissionLevel,
     procedure::{Procedure, ProcedureStage},
     repo::Repo,
+    resource::ResourceQuery,
     stack::Stack,
     update::Update,
     user::procedure_user,
@@ -33,7 +34,10 @@ use crate::{
   state::{all_resources_cache, db_client},
 };
 
-use super::update::{init_execution_update, update_update};
+use super::{
+  query::get_all_tags,
+  update::{init_execution_update, update_update},
+};
 
 pub async fn execute_procedure(
   procedure: &Procedure,
@@ -114,6 +118,7 @@ async fn execute_procedure_stage(
       Execution::BatchRunAction(exec) => {
         extend_batch_exection::<BatchRunAction>(
           &exec.pattern,
+          &exec.tags,
           &mut executions,
         )
         .await?;
@@ -121,6 +126,7 @@ async fn execute_procedure_stage(
       Execution::BatchRunProcedure(exec) => {
         extend_batch_exection::<BatchRunProcedure>(
           &exec.pattern,
+          &exec.tags,
           &mut executions,
         )
         .await?;
@@ -128,6 +134,7 @@ async fn execute_procedure_stage(
       Execution::BatchRunBuild(exec) => {
         extend_batch_exection::<BatchRunBuild>(
           &exec.pattern,
+          &exec.tags,
           &mut executions,
         )
         .await?;
@@ -135,6 +142,7 @@ async fn execute_procedure_stage(
       Execution::BatchCloneRepo(exec) => {
         extend_batch_exection::<BatchCloneRepo>(
           &exec.pattern,
+          &exec.tags,
           &mut executions,
         )
         .await?;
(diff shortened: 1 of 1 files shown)


mvanhorn:
Closing -- this is redundant now. The same fix landed upstream in 2.3.3 (#1605): `extend_batch_execution` takes a `tags` argument and filters through `get_all_tags`, and the batch expansion passes `&exec.tags` through, which is exactly what this PR was preserving. The upstream version also folds the per-variant match arms into an `expand_batch_executions!` macro, so rebasing this branch would leave an empty diff.

Thanks for the fix on your side.

Issue #1139, opened 2026-02-05
used_gb: wrong values


oke24:
Hello, there is a problem with memory alerts.

Description
Alarms are generated at minute intervals due to alleged memory overloads. 
The alarm is triggered by extremely high memory usage at “used_gb”:
```
{
  “id”: “6945676c402a1164dd4056d7”,
  “name”: “proxmox2-lxc-office”,
  “region”: null,
  “used_gb”: 17179869183.999798,
  “total_gb”: 8
}
```
The alarm is canceled again in the same minute:
```
{
  “id”: “6945676c402a1164dd4056d7”,
  “name”: “proxmox2-lxc-office”,
  “region”: null,
  “used_gb”: 0.0,
  “total_gb”: 8
}
```

Both values are incorrect. 
Komodo shows in the server view:

<img width="466" height="184" alt="Image" src="https://github.com/user-attachments/assets/872182a4-3c34-4d03-a683-65aea0084165" />

My configuration
Komodo version 1.19.5

<img width="311" height="243" alt="Image" src="https://github.com/user-attachments/assets/1159eeb5-6843-453e-a05a-d43b18c7d19a" />

oke24:
I like to add another view from Komodo which shows the wrong vaules for memory (gb_used):

<img width="1265" height="353" alt="Image" src="https://github.com/user-attachments/assets/7595be12-6d77-4201-b5f1-701117dc21e2" />

namanvashistha:
Same symptom here on an unprivileged Proxmox LXC (1 GiB), Core and Periphery on the `2` tag, Periphery running in Docker. I dug into why, and there are actually **two independent scope mismatches** stacked on top of each other. Posting the details since this thread has the symptom but not the mechanism.

`bin/periphery/src/stats/mem.rs` computes:

```rust
let raw_used = meminfo.total.saturating_sub(meminfo.available);
used: raw_used.saturating_sub(zfs_arc)
```

with `zfs_arc` read unconditionally from `/proc/spl/kstat/zfs/arcstats`, defaulting to `0` only when the file can't be read.

### 1. The ARC is host-scoped, the totals are guest-scoped

`/proc/meminfo` is virtualised by lxcfs, so `MemTotal`/`MemAvailable` describe the container. `/proc/spl/kstat/zfs/arcstats` is **not** virtualised, so it reports the Proxmox host's ARC. On my box:

```
# in the LXC
MemTotal:        1048576 kB      (1 GiB)
[...]

Issue #125, opened 2024-10-14
Inaccurate RAM usage detection


Chloe-ko:
Hello,

first off I want to start this off with saying I absolutely love Komodo, it's by far the best Docker(-Compose) Deployment tool I've ever used! Thank you for this software!

Now to the issue, Komodo is persistently and non-stop giving me this Alert:
<img width="529" alt="firefox_jTM2U6ZByI" src="https://github.com/user-attachments/assets/df742438-7ce9-474a-83b0-77a7e1f8a7e9">

I knew this was wrong and that I am not using this much memory, so I went to investigate.
To my surprise `free -h` actually shows the same usage!
<img width="569" alt="WindowsTerminal_zsEmIoDCvw" src="https://github.com/user-attachments/assets/d21ab8c8-7206-425b-8467-c01cb156a3e8">
Which at first had me worried, but then following this up by checking with htop and btop, they suddenly reported different (and more within reason) usages:
<img width="511" alt="ApplicationFrameHost_4vgqZXAHEi" src="https://github.com/user-attachments/assets/cbb58a85-4ce0-4dc0-86cf-69dd10e5fa2b">

Following this I went to investigate and eventually found the reason for this discrepancy: ZFS, the file system
ZFS is very memory-intensive as it reaches higher capacities (and I have a total of 82TB of storage using ZFS) as it uses what's called ZFS ARC. This is a cache that is kept in-memory.

Just like normal Linux, ZFS ARC separates it's RAM usage into "Used" and "Available" - Used RAM being Memory that it *actually* needs, while Available Memory is Memory that is being used but ZFS can release to the user at any given time if needed.
(I'll assume you already know this topic, but if not, here's a small tl;dr: https://www.linuxatemyram.com/)

I confirmed this using btop which actually has a setting for this:
<img width="549" alt="WindowsTerminal_dhZYO4d5Jz" src="https://github.com/user-attachments/assets/324c622d-9710-406d-8a3e-4b7ba6d5b610">

While this is set to true, btop shows the "correct" RAM Usage of 64GB Used, 61GB Available.
Setting this to false makes btop reflect what Komodo and `free-h` show: 119GB used, ~6GB available.

I double checked and with my other servers, where ZFS is not in use, Komodo correctly only considers the *used* RAM used, so this is confirmed to be a ZFS specific issue.

It'd be great if this behavior could be adjusted as otherwise I have an alert active 24/7, thank you :)

mbecker20:
[...]

PR #1635, opened 2026-09-16
periphery: ignore host ZFS ARC when /proc/meminfo comes from lxcfs

phedoreanu:
## Problem

Since #1489 (shipped in 2.3.0), Periphery subtracts the ZFS ARC size from used memory. Inside an LXC container on a ZFS Proxmox host that goes wrong: lxcfs serves a cgroup-scoped `/proc/meminfo`, but `/proc/spl/kstat/zfs/arcstats` is not virtualized and still shows the host's ARC. The subtraction saturates to zero, so every LXC guest smaller than the host's ARC reports `0.00 GB` used.

This is the mechanism described in https://github.com/moghtech/komodo/issues/1139#issuecomment-5280626418. On my fleet, four LXC containers (1 to 2 GiB each) on a host with a 6.3 GiB ARC have reported `mem_used_gb: 0` in every stats record since they picked up 2.3.x:

```
$ awk '$1=="size"{print $3/2^30 " GiB"}' /proc/spl/kstat/zfs/arcstats   # inside a 1.5 GiB CT
6.27 GiB
$ grep ' /proc/meminfo ' /proc/self/mountinfo
1206 1149 0:61 /proc/meminfo /proc/meminfo rw,nosuid,nodev,relatime shared:593 master:81 - fuse.lxcfs lxcfs rw,user_id=0,group_id=0,allow_other
```

## Fix

Check `/proc/self/mountinfo` for a FUSE filesystem mounted on `/proc/meminfo`. When that is the case the meminfo numbers are container-scoped and the ARC cannot be part of them, so the ARC is treated as zero (both for `used` and for the reported `mem_zfs_arc_gb`).

I preferred this over the `arc >= total` clamp suggested in the issue because the clamp still subtracts the full host ARC from any container whose limit is larger than the ARC (for example an 8 GiB container on a host with a 6 GiB ARC), which underreports instead of zeroing. Bare-metal hosts and Periphery running in Docker directly on a ZFS host see the real `/proc/meminfo` and keep the ARC handling from #1489 unchanged.

The second scope mismatch from that comment (Periphery in Docker inside an LXC seeing its own cgroup) is a separate problem and not addressed here.

## Testing

- Unit tests for the mountinfo parsing (lxcfs line vs. plain procfs) and for `parse_zfs_arc_size`.
- `cargo test -p komodo_periphery stats::mem` and `cargo fmt --all -- --check` pass.

Refs #1139, #125.


phedoreanu:
Verified on a live host. Proxmox 9 (kernel 7.0.14-17-pve, lxcfs 7.0.0-pve1, rpool on ZFS with a 6.28 GiB ARC), unprivileged Ubuntu LXC with a 1.5 GiB limit, Periphery as a systemd binary (not in Docker). Consecutive stats records for that server in Core's `Stats` collection, 15 s apart, with the patched Periphery starting between the second and the first:

```
2026-09-16T11:17:45Z  mem_used_gb=0.000  mem_buff_cache_gb=0.279  mem_zfs_arc_gb=6.278  mem_total_gb=1.5   (stock 2.3.3)
2026-09-16T11:18:00Z  mem_used_gb=0.000  mem_buff_cache_gb=0.279  mem_zfs_arc_gb=6.278  mem_total_gb=1.5   (stock 2.3.3)
2026-09-16T11:18:15Z  mem_used_gb=0.698  mem_buff_cache_gb=0.281  mem_zfs_arc_gb=0.000  mem_total_gb=1.5   (this patch)
```

`free -m` inside the container reports 681 MiB used at the same time, so the number is now the container's real usage.

